PR #6, opened 2017-10-31
Fixed module list links

jin:


allevato:
Nice—I didn't realize Github automatically fixed up links like that.

PR #8, opened 2017-10-31
Skylark library

Capstan:
Adding a `skylark_library` target for rules files suitable for use in Python-interpretation-based unit tests, as well as a parent target for [Skydoc doc generation](https://skydoc.bazel.build/)'s `skydoc_library`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,6 @@ Steps to add a module to Skylib:
 
 ## `skylark_library`
 
-The `skylark_library.bzl` rule allows aggregation of Skylark
-rules for use in Python-interpreter-based unit tests and documentation
-generation.
+The `skylark_library.bzl` rule can be used to aggregate a set of
+Skylark files and its dependencies for use in test targets and
+documentation generation.
--- a/README.md
+++ b/README.md
@@ -70,3 +70,9 @@ Steps to add a module to Skylib:
    ```
 
 1. Add unit tests for your module in the `tests` directory.
+
+## `skylark_library`
+
+The `skylark_library.bzl` rule allows aggregation of Skylark
+rules for use in Python-interpreter-based unit tests and documentation
+generation.
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -1,6 +1,6 @@
 licenses(["notice"])
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//:__pkg__"])
 
 load("//:skylark_library.bzl", "skylark_library")
 
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -1,8 +1,53 @@
 licenses(["notice"])
 
+package(default_visibility = ["//visibility:public"])
+
+load("//:skylark_library.bzl", "skylark_library")
+
 filegroup(
     name = "test_deps",
     srcs = ["BUILD"] + glob(["*.bzl"]),
     testonly = True,
     visibility = ["//visibility:public"],
+)
+
+skylark_library(
+    name = "collections",
+    srcs = ["collections.bzl"],
+)
+
+skylark_library(
+    name = "dicts",
+    srcs = ["dicts.bzl"],
+)
+
+skylark_library(
+    name = "paths",
+    srcs = ["paths.bzl"],
+)
+
+skylark_library(
+    name = "selects",
+    srcs = ["selects.bzl"],
+)
+
+skylark_library(
+    name = "sets",
+    srcs = ["sets.bzl"],
+)
+
+skylark_library(
+    name = "shell",
+    srcs = ["shell.bzl"],
+)
+
+skylark_library(
+    name = "structs",
+    srcs = ["structs.bzl"],
+)
+
+skylark_library(
+    name = "unittest",
+    srcs = ["unittest.bzl"],
+    deps = [":sets"],
 )
--- a/skylark_library.bzl
+++ b/skylark_library.bzl
@@ -1,0 +1,103 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -70,3 +70,9 @@ Steps to add a module to Skylib:
    ```
 
 1. Add unit tests for your module in the `tests` directory.
+
+## `skylark_library`
+
+The `skylark_library.bzl` rule allows aggregation of Skylark
+rules for use in Python-interpreter-based unit tests and documentation
```
allevato: I don't think we want to focus on the Python-interpreter part here. The important part about `skylark_library` is that it creates a proper deps tree for a library so that _any_ test target (for example, a shell test that creates a temporary workspace) can pull in the .bzl files it depends on and ensure that all the dependencies are also available.

Suggestion: "The \`skylark_library\` rule can be used to aggregate a set of Skylark files and its dependencies for use in test targets and documentation generation."

skylark_library.bzl
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Skylib module containing a library rule for aggregating rules files."""
+
+SkylarkLibraryInfo = provider()
```
allevato: Please add `doc` and `fields` to this with the relevant documentation.

skylark_library.bzl
```diff
@@ -0,0 +1,103 @@
...
+def _skylark_library_impl(ctx):
+  all_files = depset(ctx.files.srcs, order="postorder")
+  for dep in ctx.attr.deps:
+    all_files += dep.files
+  return [
+      # All dependent files should be listed in both `files` and in `runfiles`;
+      # this ensures that a `library()` can be referenced as `data` from a
+      # separate program, or from `tools` of a genrule().
+      DefaultInfo(
+          files = all_files,
```
allevato: Please remove spaces around `=` in kwargs lists (here, below, and in the `rule` call).

skylark_library.bzl
```diff
@@ -0,0 +1,103 @@
...
+
+SkylarkLibraryInfo = provider()
+
+def _skylark_library_impl(ctx):
+  all_files = depset(ctx.files.srcs, order="postorder")
+  for dep in ctx.attr.deps:
+    all_files += dep.files
+  return [
+      # All dependent files should be listed in both `files` and in `runfiles`;
+      # this ensures that a `library()` can be referenced as `data` from a
```
allevato: s/library()/skylark_library/

lib/BUILD
```diff
@@ -1,8 +1,53 @@
 licenses(["notice"])
 
+package(default_visibility = ["//visibility:public"])
```
allevato: The visibility of the individual `skylark_library` targets in this directory should be private so that users can only load them through `lib.bzl`. If you change that, does everything still work (`/...` builds and tests pass)?

README.md
Capstan: Done.

lib/BUILD
Capstan: Yes; done.

skylark_library.bzl
Capstan: Done. PTAL.

skylark_library.bzl
Capstan: Done.

skylark_library.bzl
Capstan: Done-ish. Didn't quite know what you were referring to with the `rule` call; there's documentation about how to use it in a `BUILD` file which has different style guides.

skylark_library.bzl
```diff
@@ -0,0 +1,103 @@
...
+skylark_library = rule(
+    implementation = _skylark_library_impl,
+    attrs = {
+        "srcs": attr.label_list(
+            allow_files=[".bzl"],
+        ),
+        "deps": attr.label_list(
+            providers=[
+                [SkylarkLibraryInfo],
+                ["files"],
```
Capstan: The `"files"` provider is a break-glass for the case where we need to include Skylark rules from `baselbuild/bazel`, e.g., https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/hash/hash.bzl, and they wouldn't want to break layering to include from bazel-skylib. Is there a better way to limit these to `.bzl` files, or even better, certain whitelisted targets?

skylark_library.bzl
allevato: In the call to the `rule` function at the end of the file, where the rule is defined, those are kwargs so I don't put any spaces around the `=`. The only places where I use spaces around `=` are in BUILD files when creating a target based on a rule, or in macros where a rule is being invoked.

It's an unfortunate style inconsistency, I'll admit.

skylark_library.bzl
Capstan: Exactly, and you're entirely correct. So, is there a filegroup-provided provider that can apply these filters, or does `attr.label_list`'s `allow_files` suffice here? Does that just filter the files in said group?

allevato: I'm not aware of anything off the top of my head.

One possibility I can think of is, in the impl function for `skylark_library`, just exclude any files that don't end in `.bzl`. It's still hacky, but it at least prevents the case we're trying to avoid. WDYT?

Capstan: Looks like with `allow_files` and use of an `exports_files` in `tools/build_defs/hash/BUILD` (instead of the existing `:srcs` target) would suffice here, and can remove this break-glass.

skylark_library.bzl
Capstan: Done.

skylark_library.bzl
Capstan: Removed.

dslomov:
awesomesauce

PR #15, opened 2018-01-18
Fix paths.is_absolute on windows

ianthehat:


lib/paths.bzl
```diff
@@ -65,7 +65,7 @@ def _is_absolute(path):
   Returns:
     `True` if `path` is an absolute path.
   """
-  return path.startswith("/")
+  return path.startswith("/") or (len(path)>2 and path[1] == ":")
```
allevato: Please put spaces around the `>` operator.

Can you also add a comment that the length check is there to catch `C:/...` but not `C:`? The [Bazel `PathFragment` source](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/vfs/PathFragment.java#L635-L640) has a similar comment, so that would be helpful to remember that subtle detail.

allevato: Looks like I was too slow!

PR #29, opened 2018-03-07
Don't special case current directory relative paths.

thomasvl:


tests/paths_tests.bzl
```diff
@@ -195,6 +195,10 @@ def _relativize_test(ctx):
   # Try a case where a parent directory is normalized away.
   asserts.equals(env, "baz", paths.relativize("foo/bar/../baz", "foo"))
 
+  # Relative paths work, as long as they common.
```
sergiocampama: they share a common root?

thomasvl: Fixed

PR #41, opened 2018-05-02
Add strings lib with strip_margin utility

andyscott:
The primary use case for the new utility is to be able to write inline multiline strings with indentation that matches the rest of the file.

Example from some rules I've been working on:
```
def _some_rule_impl(ctx):
    # ...
    ctx.actions.run_shell(
        inputs = inputs,
        outputs = [output_jar],
        command = strip_margin(
            """
            |set -eo pipefail
            |
            |mkdir tmp/classes
            |
            |{java} \\
            |  -cp {compiler_classpath} \\
            |  scala.tools.nsc.Main \\
            |  -cp {compile_classpath} \\
            |  -d tmp/classes \\
            |  {srcs}
            |
            |{jar_creator} {output_jar} tmp/classes 2> /dev/null
            |""".format(
                jar = jar.path,
                java = java.path,
                jar_creator = jar_creator.path,
                compiler_classpath = compiler_classpath_str,
                compile_classpath = compile_classpath_str,
                srcs = srcs_string,
                src_jars = src_jars_string,
                output_jar = output_jar.path,
            ),
        ),
    )
```

### DIFF
--- a/lib.bzl
+++ b/lib.bzl
@@ -22,6 +22,7 @@ load("//lib:paths.bzl", _paths="paths")
 load("//lib:selects.bzl", _selects="selects")
 load("//lib:sets.bzl", _sets="sets")
 load("//lib:shell.bzl", _shell="shell")
+load("//lib:strings.bzl", _strings="strings")
 load("//lib:structs.bzl", _structs="structs")
 load("//lib:versions.bzl", _versions="versions")
 
@@ -38,6 +39,7 @@ paths = _paths
 selects = _selects
 sets = _sets
 shell = _shell
+strings = _strings
 structs = _structs
 versions = _versions
 
--- a/lib/strings.bzl
+++ b/lib/strings.bzl
@@ -1,0 +1,48 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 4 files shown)


brandjon:
I recommend against the '|' formatting to indicate where the line begins. It's cumbersome to type and not what people write in practice. If the user just wants a multiline comment like a docstring, and only cares what it looks like syntactically in source and not what string value it translates to, they'll indent without the '|'. (Markdown uses the same philosophy -- design the syntax around what people already do.)

For prior art, see Python's textwrap.dedent(): https://docs.python.org/3/library/textwrap.html#textwrap.dedent

I recommend this approach -- just remove the greatest common leading whitespace prefix of all lines. Alternatively, exempt the first line from consideration, so you can write

    # (left margin is here)
        """first line
        second line
        third line
        """

and the four leading spaces from the second and third lines will be trimmed even though the first line has no leading whitespace.

As an aside, note that multiline trimming is needed for docstrings, both of the form

    def foo():
        """stuff about
        the foo function
        """
        ...

and

    myrule = rule(
        ...
        doc = """stuff about
            the myrule rule
            """

but since tooling may expect a string literal rather than a string expression (especially in the first case), such trimming would probably be done by doc tooling rather than by an explicit function the user calls.

brandjon:
Wha, didn't mean to close...

andyscott:
@brandjon Sounds good to me. I'm not terribly picky-- this particular implementation is derived from Scala's `stripMargin` function. However, it seems like a `dedent` derived solution would be more appropriate.

I'll update this PR with a dedent method to replace strip_margin.

brandjon:
The algorithm in the PEP you mentioned sounds great for docstring processing, and a doc tool could use it. But as a general utility function, users might be surprised that it eats leading and trailing blank lines (would it affect scripts like in the above example?). Or not. So long as it's clearly specified, I could go either way.

c-parsons:
By looking at the log here, I believe this review was hanging on the promise of additional changes: "I'll update this PR with a dedent method to replace strip_margin."

I know this PR is very old now.

Should we close this?

aiuto:
Friendly ping to @andyscott

tetromino:
Closing - we'd prefer something like python's `dedent`.

PR #42, opened 2018-05-04
Add support for repr/str to new_sets

dmaclach:


lib/new_sets.bzl
```diff
@@ -220,6 +220,17 @@ def _length(s):
...
   return len(s._values)
 
+def _repr(s):
+  """Returns a string value representing the set.
+
+  Args:
+    s: A set, as returned by `sets.make()`.
+
+  Returns:
+    An string representing the set.
```
allevato: s/An string/A string/

lib/new_sets.bzl
```diff
@@ -235,4 +246,6 @@ sets = struct(
   difference = _difference,
   length = _length,
   remove = _remove,
+  repr = _repr,
+  str = _repr,
```
allevato: To future self (and other readers): According to a [comment in the `str` implementation](https://github.com/bazelbuild/bazel/blob/e2d1a55607362804f07d3234d514f39cb5eed45a/src/main/java/com/google/devtools/build/lib/syntax/Printer.java#L335-L341), the output of `str` and `repr` only differ for top-level strings and labels. Since the values here will always be in a list, they should be identical.

If this behavior ever changes, we'll need to revisit this equivalence.

PR #43, opened 2018-05-08
Add type checking functions.

ttsugriy:
Even though it's not great to use type checks, they are frequently useful for
checking input types of macros.

Because there is no standard way of checking types, at least 2 types of checks
are used:
- `type(foo) == type([])`
- `type(foo) == "list"`

The first option is not very readable and the second option seem to be relying
on an Bazel implementation detail. Encapsulating type checks into this library
enables consistent and easy to understand type checking without explicitly
relying on implementation details.

There are still more checks to add, but this change creates a scaffold.

### DIFF
--- a/lib/types.bzl
+++ b/lib/types.bzl
@@ -1,0 +1,135 @@
+# Copyright 2018 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Skylib module containing functions checking types."""
+
+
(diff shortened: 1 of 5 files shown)


ttsugriy:
awesome, thanks for such a supersonic review @allevato! I'll update the PR with copyright update and the rest of type checks!

ttsugriy:
added all missing type checks and fixed copyright, please take a another look. Thanks!

lib/types.bzl
```diff
@@ -0,0 +1,135 @@
...
+_a_tuple = ()
+_an_int = 1
+
+
+def _a_function():
+  pass
+
+
+def _is_list(v):
+    """Returns True if v is an instance of the list.
```
allevato: Super-nit: Here and in the `Returns` clause, can you say "a list", etc. instead of "the list"? (Throughout the docs for each of these functions.)

ttsugriy:
thanks for the review @allevato! I was going back and forth between `a` and `the` so thanks for putting an end to my internal battle :) I updated all usages to use `a` instead of `the`

laurentlb:
Thanks! I also wanted a library like this :)

brandjon:
LGTM

allevato:
Thanks for your contribution, @ttsugriy!  (Laurent beat me to merging it by 30 seconds... 🙂)

ttsugriy:
thanks a lot for such a thorough and supersonic review!

Issue #30, opened 2018-03-13
paths.relativize doesn't support a path not fully under start


thomasvl:
The docs for `paths.relativize` compares itself to `os.path.relpath`, but the current implementation requires `path` be fully under `start`.

In theory it could also work for things that have a common prefix:
* `paths.relativize("foo/bar/baz", "foo/mumble")` could return `"../bar/baz"`
* `paths.relativize("foo/bar", "foo/mumble/baz")` could return `"../../bar"`
* `paths.relativize("foo/bar/baz", "mumble/grumble")` could return `"../../foo/bar/baz"`

Right now these cases `fail`, so adding the support shouldn't break anyone.  If callers want something fully nested, they could check the result to ensure it doesn't start with `..`.

PR #44, opened 2018-05-15
Add ‘paths.relative_path’ function

mrkkrp:
This is a function that we found ourselves needing sometimes.


lib/paths.bzl
```diff
@@ -195,6 +195,30 @@ def _relativize(path, start):
   result_segments = segments[-length:]
   return "/".join(result_segments)
 
+def _relative_path(target, start):
+  """Returns a relative path to `target` from `start`.
+
+  Args:
+    target: path that we want to get relative path to.
+    start: path to directory from which we are starting.
```
thomasvl: I guess the assumption is both paths are rooted at the same place to start with?

Should it check for getting a relative path in either arg?

This is pretty close to relativize, the docs are going to need to call out exactly what is different and why (assuming it is worth having both).

tests/paths_tests.bzl
```diff
@@ -205,6 +205,28 @@ def _relativize_test(ctx):
...
+  asserts.equals(env, ".", paths.relative_path("/", "/"))
+
+  # When paths match.
+  asserts.equals(env, ".", paths.relative_path("foo/bar", "foo/bar"))
+
+  # Normal functioning.
+  asserts.equals(env, "../baz", paths.relative_path("/foo/bar/baz", "/foo/bar/quux"))
+  asserts.equals(env, "../../ba0/baz", paths.relative_path("/foo/ba0/baz", "/foo/ba1/quux"))
+  asserts.equals(env, "../../foo", paths.relative_path("foo", "bar/baz"))
+  asserts.equals(env, "../../foo", paths.relative_path("/foo", "/bar/baz"))
```
thomasvl: The tests seem to almost imply if one is absolute, the both have to be absolute, but the code doesn't ensure that.

lib/paths.bzl
mrkkrp: The function is completely different. `paths.relativize` requires the path you want to relativize to contain the other argument as its prefix. Here we do a more general thing: we tell how to get to `target` path starting from `start` path. `paths.relativize` hasn't been very useful to us, while the new function is quite handy.

allevato: A better path forward then would be to simply extend the existing function to handle that scenario. Adding a new function with a similar name will simply lead to confusion.

mrkkrp: Sometimes you don't want to use absolute paths, e.g. when writing rpaths.

mboes:
Related: https://github.com/bazelbuild/bazel-skylib/issues/30.

aiuto:
This seems to have gone stale? Is it going to move forward or should we close it.

tetromino:
Is there a chance of this getting reviewed and merged? I just wrote an almost identical function in starlark for some rules this afternoon because it didn't find it in Skylib - would have been nice if it were available :)

lib/paths.bzl
gonzojive: I suggest using relativize and adding a keyword argument to specify this .. behavior is desired. Otherwise it might break people who assume relativize fails under certain scenarios.

PR #52, opened 2018-07-27
initial checkin of skydoc skylark rule

c-parsons:


laurentlb:
Why bazel-skylib and not https://github.com/bazelbuild/skydoc?

c-parsons:
I suppose skydoc would be okay. Our cross-repo dependencies would then be:

skydoc -> bazel_skylib -> io_bazel
skydoc tests, in io_bazel -> skydoc -> bazel_skylib

laurentlb:
Skydoc repository depends on both Skylib and Bazelbuild repositories.
This should be the same for the tests (unless I missed something?)

thomasvl:
@c-parsons is this being dropped?

c-parsons:
Yes, I'd like to drop this. I have taken the alternative approach of incorporating into the skydoc directory, which I'm waiting on approval for. https://github.com/bazelbuild/skydoc/pull/105
I was dragging my feet on dropping this in case my reviewers happened to change their mind, but I doubt they will 0:)

PR #54, opened 2018-08-24
All incompatible changes

jjudd:
This enables --all_incompatible_changes with Bazel 0.16. With this change, when things are deprecated in the future, builds will fail, and the deprecated features can be changed before they are actually removed from Bazel/Starlark.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+common --all_incompatible_changes
--- a/lib/new_sets.bzl
+++ b/lib/new_sets.bzl
@@ -34,6 +34,8 @@ def _make(elements = None):
       A set containing the passed in values.
     """
     elements = elements if elements else []
+    if type(elements) == "depset":
+        elements = elements.to_list()
     return struct(_values = {e: None for e in elements})
 
 def _copy(s):
--- a/lib/sets.bzl
+++ b/lib/sets.bzl
@@ -40,6 +40,16 @@ def _precondition_only_sets_or_lists(*args):
             fail("Expected arguments to be depset or list, but found type %s: %r" %
                  (t, a))
 
+def _convert_to_list(a):
+    """If the given value is a depset, convert it to a list. Otherwise do nothing.
+
+    Args:
+      *args: The value to be converted
+    """
+    if type(a) == "depset":
+        a = a.to_list()
+    return a
+
 def _is_equal(a, b):
     """Returns whether two sets are equal.
 
@@ -51,7 +61,7 @@ def _is_equal(a, b):
       True if `a` is equal to `b`, False otherwise.
     """
     _precondition_only_sets_or_lists(a, b)
-    return sorted(depset(a)) == sorted(depset(b))
+    return sorted(depset(a).to_list()) == sorted(depset(b).to_list())
 
 def _is_subset(a, b):
     """Returns whether `a` is a subset of `b`.
@@ -64,8 +74,8 @@ def _is_subset(a, b):
       True if `a` is a subset of `b`, False otherwise.
     """
     _precondition_only_sets_or_lists(a, b)
-    for e in a:
-        if e not in b:
+    for e in _convert_to_list(a):
+        if e not in _convert_to_list(b):
             return False
     return True
 
@@ -82,8 +92,8 @@ def _disjoint(a, b):
       True if `a` and `b` are disjoint, False otherwise.
     """
     _precondition_only_sets_or_lists(a, b)
-    for e in a:
-        if e in b:
+    for e in _convert_to_list(a):
+        if e in _convert_to_list(b):
             return False
     return True
 
@@ -98,14 +108,14 @@ def _intersection(a, b):
       A set containing the elements that are in both `a` and `b`.
     """
     _precondition_only_sets_or_lists(a, b)
-    return depset([e for e in a if e in b])
+    return depset([e for e in _convert_to_list(a) if e in _convert_to_list(b)])
 
 def _union(*args):
     """Returns the union of several sets.
 
     Args:
       *args: An arbitrary number of sets or lists.
-
+      *
     Returns:
       The set union of all sets or lists in `*args`.
     """
@@ -124,7 +134,7 @@ def _difference(a, b):
       A set containing the elements that are in `a` but not in `b`.
     """
     _precondition_only_sets_or_lists(a, b)
-    return depset([e for e in a if e not in b])
+    return depset([e for e in _convert_to_list(a) if e not in _convert_to_list(b)])
 
 sets = struct(
     difference = _difference,
--- a/lib/sets.bzl
+++ b/lib/sets.bzl
@@ -44,7 +44,7 @@ def _convert_to_list(a):
     """If the given value is a depset, convert it to a list. Otherwise do nothing.
 
     Args:
-      *args: The value to be converted
+      a: The value to be converted
     """
     if type(a) == "depset":
         a = a.to_list()
@@ -115,7 +115,6 @@ def _union(*args):
 
     Args:
       *args: An arbitrary number of sets or lists.
-      *
     Returns:
       The set union of all sets or lists in `*args`.
     """
--- a/skylark_library.bzl
+++ b/skylark_library.bzl
@@ -32,9 +32,8 @@ def _skylark_library_impl(ctx):
         # a separate program, or from `tools` of a genrule().
         DefaultInfo(
             files = all_files,
-            runfiles = ctx.runfiles(files = list(all_files)),
+            runfiles = ctx.runfiles(files = all_files.to_list()),
         ),
-
         # We also define our own provider struct, for aggregation and testing.
         SkylarkLibraryInfo(
             srcs = ctx.files.srcs,
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -166,10 +166,10 @@ def _end(env):
         "EOF",
         "exit %d" % len(env.failures),
     ])
-    env.ctx.file_action(
+    env.ctx.actions.write(
         output = env.ctx.outputs.executable,
         content = cmd,
-        executable = True,
+        is_executable = True,
     )
 
 def _fail(env, msg):


skylark_library.bzl
```diff
@@ -32,9 +32,8 @@ def _skylark_library_impl(ctx):
         # a separate program, or from `tools` of a genrule().
         DefaultInfo(
             files = all_files,
-            runfiles = ctx.runfiles(files = list(all_files)),
+            runfiles = ctx.runfiles(files = all_files.to_list()),
```
laurentlb: I don't think `to_list` is needed here.

lib/sets.bzl
```diff
@@ -40,6 +40,16 @@ def _precondition_only_sets_or_lists(*args):
...
 
+def _convert_to_list(a):
+    """If the given value is a depset, convert it to a list. Otherwise do nothing.
+
+    Args:
+      *args: The value to be converted
+    """
+    if type(a) == "depset":
+        a = a.to_list()
+    return a
```
laurentlb: This file has a broken design (it shouldn't have used `depset` in the first place).
Your fix looks reasonable. An alternative is to keep the code as is. It will break in the future, but this will force users to update to `new_sets` instead.

lib/new_sets.bzl
```diff
@@ -34,6 +34,8 @@ def _make(elements = None):
       A set containing the passed in values.
     """
     elements = elements if elements else []
+    if type(elements) == "depset":
+        elements = elements.to_list()
```
laurentlb: I recommend against it: https://docs.bazel.build/versions/master/skylark/performance.html#never-call-depset-to_list
Flattening or iterating over a `depset` should be explicit.

thomasvl: @laurentlb the Depsets page says otherwise:

https://g3doc.corp.google.com/devtools/blaze/rules/g3doc/depsets.md#upcoming-changes

> * Depset contents should be retrieved using to_list(), not by iterating over the depset itself. Direct iteration over depsets is deprecated and will be removed. I.e., don't use list(...), sorted(...), or other functions expecting an iterable, on depsets.

.bazelrc
```diff
@@ -0,0 +1 @@
+common --all_incompatible_changes
```
brandjon: To be clear, adding a .bazelrc to this repo shouldn't have any effect on users who import this repo to their own workspace? I can't imagine that it would, but if it did that'd be a show-stopper since we would be breaking downstream users.

lib/new_sets.bzl
brandjon: The page you're citing is talking about how implicit iteration over depsets is deprecated, and the preferred alternative is to do it explicitly with `to_list()`. In this case, a library function is wrapping the call to `to_list()`, so Laurent is concerned that it's effectively being made implicit once again, at least from the POV of the caller of `sets.make()`.

(You could argue that if you're asking to make a hashset out of a depset, that in itself is a form of explicit flattening. Not sure where I stand on that.)

In the previous code, users can run with `--all_incompatible_changes` to know whether or not they have any call sites that do `sets.make(mydepset)` when they really should've been rewritten to not flatten depsets, or at least marked explicitly with `sets.make(mydepset.to_list())`. With this change, `sets.make(mydepset)` will always be allowed with or without `--all_incompatible_changes`.

On the other hand, if we revert this change, the `sets.make(mydepset)` usage will continue to be an error if and only if the incompatible flag is given. In the interest of preserving semantics in this PR I think that's what we should do.

lib/sets.bzl
allevato: IMO, I'd also prefer we leave it broken to force migration, since it's effectively deprecated now.

brandjon: Is there a plan to rename `new_sets` to `sets`? If so, is that simply going to happen after `sets` is removed?

Ideally we'd break this module with an explicit `fail()` that is sensitive to the Bazel version and/or the breaking `--incompatible_*` flags. Then we can produce a clean error message for the user, instead of frustrating them and leaving it to them to figure out which dependency broke and why. I don't think we expose the necessary info to Starlark at the moment though.

lib/sets.bzl
```diff
@@ -98,14 +108,14 @@ def _intersection(a, b):
...
-    return depset([e for e in a if e in b])
+    return depset([e for e in _convert_to_list(a) if e in _convert_to_list(b)])
 
 def _union(*args):
     """Returns the union of several sets.
 
     Args:
       *args: An arbitrary number of sets or lists.
-
+      *
```
brandjon: This line seems misplaced.

.bazelrc
jjudd: That is my understanding, yes. This will not affect users who import Skylib into their workspace.

The advantage of doing this is for users who do use `--all_incompatible_changes` in their workspace and import Skylib. If Skylib does not build with `--all_incompatible_changes`, then their repo will fail to build. That's the motivation for this change.

lib/new_sets.bzl
jjudd: I don't have access to the g3doc page, so please pardon me I'm not following the conversation exactly :D 

It would be nice if this function was compatible with `--all_incompatible_changes`, regardless of the arguments passed to it. If we want to make this code not accept depsets, wouldn't it be better to fail when a depset is passed to it and still be compatible with `--all_incompatible_changes`?

laurentlb: The g3doc page is here: https://docs.bazel.build/versions/master/skylark/depsets.html#upcoming-changes

lib/sets.bzl
jjudd: I'd prefer that the file be compatible with `--all_incompatible_changes`. If we want people to migrate, then I like removing this file or making it fail based on Bazel version. The `--incompatible_*` warning messages are not terribly helpful for knowing that I need to migrate to `new_sets`. An explicit error message would make it much clearer.

lib/sets.bzl
jjudd: Thanks.  Will fix.

lib/new_sets.bzl
thomasvl: Edited/corrected my link above. Sorry, I meant to grab the external one in the first place.

skylark_library.bzl
jjudd: Removing this change with `--all_incompatible_changes` enabled produces this error on `bazel build ...`

```
ERROR: /home/james/opensource/bazel-skylib/lib/BUILD:53:1: in skylark_library rule //lib:structs: 
Traceback (most recent call last):
	File "/home/james/opensource/bazel-skylib/lib/BUILD", line 53
		skylark_library(name = 'structs')
	File "/home/james/opensource/bazel-skylib/skylark_library.bzl", line 33, in _skylark_library_impl
		DefaultInfo(files = all_files, runfiles = ctx....)))
	File "/home/james/opensource/bazel-skylib/skylark_library.bzl", line 35, in DefaultInfo
		ctx.runfiles(files = list(all_files))
	File "/home/james/opensource/bazel-skylib/skylark_library.bzl", line 35, in ctx.runfiles
		list(all_files)
type 'depset' is not iterable. Use the `to_list()` method to get a list. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check.
ERROR: Analysis of target '//lib:structs' failed; build aborted: Analysis of target '//lib:structs' failed; build aborted
```

lib/sets.bzl
laurentlb: Having an explicit error message sounds good.
I don't think it makes sense to maintain this file.

skylark_library.bzl
laurentlb: What about this?
`            runfiles = ctx.runfiles(files = all_files),`

lib/new_sets.bzl
jjudd: Thanks for updating the link! :)

thomasvl:
Do we want to wrap this up and land it so we can cut a release along with #59 and #58?

jjudd:
I'm happy to jump in and change things, but I'm not sure where we landed on the changes.

c-parsons:
It's also a little unclear on where the reviews stand (sorry, I'm not great at understanding what's unresolved in github reviews..) What questions are we blocked on?

brandjon:
From a quick reread of this thread, I think the only outstanding issue is whether to implicitly flatten depsets in `_make()` in order to avoid triggering failures when users run their calling code under `--incompatible_depset_is_not_iterable`. My opinion is that this PR should not workaround these failures at the cost of efficiency, but rather force the calling code to explicitly flatten with `.to_list()`, since that's more in line with the intent of the flag.

lib/new_sets.bzl
allevato: new_sets.bzl was introduced because we had folks internally using the broken sets.bzl implementation, "dictionary of Bools" is kind of a busted abstraction to recommend to users who need set functionality, and the module lets us expose important operations like union/intersection more cleanly than just using the raw dictionary.

c-parsons:
I'd like to ping this PR again. Are folks on this thread actively working on resolving this review?

I would really like to cut a release of bazel-skylib in the very near future, as some recent PRs are a requirement for Stardoc. I also have some upcoming changes in mind that would cause merge conflicts with this PR, so I'd really like to resolve this soon.

If this has become a lower priority to resolve, please let me know.

c-parsons:
Any updates?

c-parsons:
Ping.
It's been 3 weeks since we've gotten any updates on this PR, and it has been open since August.

I plan on cloning this PR and pushing it through myself if I don't hear anything back by Monday.

bttk:
> Ping.
> It's been 3 weeks since we've gotten any updates on this PR, and it has been open since August.
> 
> I plan on cloning this PR and pushing it through myself if I don't hear anything back by Monday.
@c-parsons 
I have addressed some incompatible changes in #67 #68 #69

c-parsons:
Thanks a bunch, @bttk !

bttk:
@c-parsons Could you also take a took at #70 ?

PR #56, opened 2018-08-24
Remove usage and suggests for lib.bzl.

thomasvl:
As more things are added, lib.bzl is an anti-pattern as the cost of loading
it actually just keeps increasing and most things will never use everything
out of it.  The pattern the should be used is to directly import the modules
one uses.


thomasvl:
It probably makes sense to add a deprecation warning to the `lib` target and kick out a message from `lib.bzl`, but I figured it we should probably any message like that at the same time we message about `set` vs. `new_set`.  So for now, this just is about the readme and the code within this project.

laurentlb:
Merged. Thanks!

PR #58, opened 2018-09-19
deprecate lib.bzl.

thomasvl:
- tag the :lib target as such.
- print() a warning when lib.bzl is loaded.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,6 +21,10 @@ filegroup(
 skylark_library(
     name = "lib",
     srcs = ["lib.bzl"],
+    deprecation = (
+        "lib.bzl will go away in the future, please directly depend on the" +
+        " module(s) needed as it is more efficient."
+    ),
     deps = [
         "//lib:collections",
         "//lib:dicts",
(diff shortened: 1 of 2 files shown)


thomasvl:
@laurentlb ping - should we move forward with this or is there something better than the `print` to use?

laurentlb:
What about `fail`? If someone updates their version of skylib, they'll get the error message and fix the code.

thomasvl:
> What about fail instead of print? If someone updates their version of skylib, they'll get the error message and fix the code.

I'm a little worried about the deps problem –

Developer uses two rules sets (A & B), both use skylib, _A_ is using `lib.bzl` and _B_ isn't.  They need a new version of _B_ which needs a newer Skylib, but they can't update because a `fail` from _A_ would break their build.  Now what?

Seems like the best bet is to fake what a `deprecation` rule attribute would do via `print`, and ship a release (or two) with that to warn folks, before moving to `fail`.

laurentlb:
Sounds good.

PR #60, opened 2018-10-18
Remove spurious newlines from fail/print messages

jmmv:
The fail() and print() primitives take strings without newlines in them
because the formatted output will already contain newlines.  Therefore,
remove them, which have been annoying me to no end when using rules_go
with a HEAD-built Bazel binary.

While doing this, also switch to using .format() consistently for all
strings, merge two related debug messages into the same print() call,
and fix grammar.


allevato:
Thanks!

PR #63, opened 2018-11-12
Update to google style

techsy730:
Update skylark_library.bzl to conform to Google's code style

Even though skylark_library.bzl is deprecated, it should still conform to Google's code style.


techsy730:
Please reject; this picked up my previous change too

skylark_library.bzl
```diff
@@ -1,10 +1,29 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
allevato: Nit: 2018

techsy730: Done

PR #68, opened 2018-11-13
Remove support for make(depset()) from new_sets

bttk:
With `--incompatible_depset_is_not_iterable` depset is no longer
iterable. Remove tests that assumed it is.

Bazel issue:
https://github.com/bazelbuild/bazel/issues/5816


bttk:
@laurentlb please take a look

PR #74, opened 2018-11-21
buildifier lint issues/validation

thomasvl:


thomasvl:
I don't think there is a way to have the webhooks report just "warnings" vs _success_/_failure_, and this means any lint issue would fail the presubmits.  The travis buildifier jobjs could be split in two, one for lint and one for format, so when looking at the travis results page it is a little easier to see when it is a formatting (easy fix) vs. linter (harder fix) issue.

PR #80, opened 2018-11-29
Fix load paths examples in README

kellycampbell:

PR #83, opened 2018-11-30
Fix example in README.md on loading from .bzl file in lib/ directly

gertvdijk:
Without this change, you would run into an error:

    ERROR: error loading package '': Extension file not found. Unable to
    load package for '@bazel_skylib//lib/paths.bzl:paths.bzl': BUILD file
    not found on package path


thomasvl:
Correct fix looks to be pending in https://github.com/bazelbuild/bazel-skylib/pull/80

PR #84, opened 2018-12-03
unittest.bzl: supports Windows

laszlocsomor:
In this commit:

- change unittest.bzl to declare a named output
  file instead of relying on the deprecated [1]
  default output name (ctx.outputs.executable).

- define a new toolchain_type and toolchain rules
  for Windows and non-Windows ("default")
  execution platforms

- register the new toolchains in workspace.bzl

- let unittest.make-created test rules require the
  new toolchain_type

- write the test output script as a Windows batch
  script or as a Shell script, depending on the
  selected toolchain

This PR enables the Bazel team to break the Bash
dependency (for test execution) on Windows so that
Bazel can run Starlark unittests with the new
Windows-native test wrapper [2] (still under
development).

[1] https://docs.bazel.build/versions/master/skylark/lib/ctx.html#outputs

[2] https://github.com/bazelbuild/bazel/issues/5508


laszlocsomor:
@katre : could you please take a look for the toolchain-related changes?

WORKSPACE
```diff
@@ -1 +1,5 @@
 workspace(name = "bazel_skylib")
+
+load(":workspace.bzl", "bazel_skylib_workspace")
```
thomasvl: Why push this into a bzl? Is it something consumers of skylib will have to do in their WORKSPACE or is this just local for the local test?  If it is local, why not just directly do the work here?

thomasvl: ps - if it is something everyone else is going to have to do, then the README likely needs to also be updated to include the info.

laszlocsomor: Yes, users of skylib have to call this macro to register the toolchains.

laszlocsomor:
Thanks, let me address the remaining comments in one more commit. I'll ping this thread when the PR is ready again.

lib/unittest.bzl
```diff
@@ -75,6 +77,13 @@ def _make(impl, attrs = None):
         attrs = attrs,
         _skylark_testable = True,
         test = True,
+        # is_host_windows uses a select(), therefore we cannot pass an output function that would
+        # compute the output file's name based on this attribute (to return "%{name}.bat" or
+        # "%{name}), so we use "%{name}.bat" on all platforms.
+        # On Windows this allows the output file to be executable. On all other platforms the file's
+        # extension is irrelevant because it's shebang line defines the interpreter.
+        outputs = {"testbin": "%{name}.bat"},
```
laurentlb: Can you avoid using `outputs =`?
Instead, declare a file (`ctx.actions.declare_file`) in the implementation function. As a bonus, it won't need the .bat extension on non-Windows platforms.

WORKSPACE
laurentlb: Is it all users of skylib? Or just the ones that use `unittest.bzl`?

lib/unittest.bzl
```diff
@@ -160,17 +163,39 @@ def _end(env):
...
       env: The test environment returned by `unittest.begin`.
     """
-    cmd = "\n".join([
-        "cat << EOF",
-        "\n".join(env.failures),
-        "EOF",
-        "exit %d" % len(env.failures),
-    ])
+
+    if env.ctx.toolchains[_TOOLCHAIN_TYPE].bazel_skylib_toolchain_info.is_exec_windows:
```
katre: In general, having rules switch on a boolean in the toolchain is a bad code smell. Specifically, this ties your toolchain to "is windows?" and "is not windows?" checks.

Instead, would it be possible to have your toolchain declare several fields, one being a file extension (".bat" vs ".sh"), and another being templates for the success and failure messages, so that this function is reduced to just a single block with no if/else check?

tests/shell_tests.bzl
```diff
@@ -96,6 +96,9 @@ def _shell_spawn_e2e_test_impl(ctx):
 shell_spawn_e2e_test = rule(
     test = True,
     implementation = _shell_spawn_e2e_test_impl,
+
+    # All unittest.suite expects the rule to have an "is_host_windows" bool attribute.
+    attrs = {"is_host_windows": attr.bool()},
```
katre: Who is setting this attribute? Putting it on targets in a BUILD file seems backwards.

lib/unittest.bzl
laszlocsomor: Done. Great suggestion, thank you!

tests/shell_tests.bzl
laszlocsomor: Done.
Was obsolete.

WORKSPACE
laszlocsomor: After @katre 's comment (https://github.com/bazelbuild/bazel-skylib/pull/84#discussion_r238336792) I changed the approach. Now unittest.bzl defines the toolchain, and maprule will define its own toolchain, so that the platform-specific logic can be moved there.

toolchains/unittest/BUILD
```diff
@@ -0,0 +1,51 @@
...
+
+unittest_toolchain(
+    name = "exec_batch",
+    failure_templ = "\n".join([
+        "@echo off",
+        "echo %s",
+        "exit /b 1",
+    ]),
+    file_ext = ".bat",
+    join_on = "\necho",
```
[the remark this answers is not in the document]
laszlocsomor: Good catch! Done.
I just tried using importing skylib as an external repo and using unittest and it seems to work.

lib/unittest.bzl
```diff
@@ -22,7 +22,40 @@ assertions used to within tests.
...
+    implementation = _unittest_toolchain_impl,
+    attrs = {
+        "file_ext": attr.string(mandatory = True),
+        "success_templ": attr.string(mandatory = True),
+        "failure_templ": attr.string(mandatory = True),
+        "join_on": attr.string(mandatory = True),
+    },
+)
+
+def register_unittest_toolchains():
```
katre: I'm a bit uncomfortable mixing macros and rules intended for a BUILD file with those intended for a WORKSPACE file. Maybe I'm just over-sensitive? In either case, add a comment that this will only work if the macro is called from WORKSPACE.

toolchains/unittest/BUILD
```diff
@@ -0,0 +1,51 @@
+load("//lib:unittest.bzl", "TOOLCHAIN_TYPE", "unittest_toolchain")
+
+toolchain_type(
+    name = "toolchain_type",
+)
+
+unittest_toolchain(
+    name = "exec_batch",
+    failure_templ = "\n".join([
```
katre: Is there a reason not to use a multiline string instead of join? If so, add a comment explaining why.

lib/unittest.bzl
laszlocsomor: Done: added comment.
I can't think of any problems now about mixing workspace and build rules in the same .bzl file. If we later discover problems, it should be easy to split them.

toolchains/unittest/BUILD
laszlocsomor: I think it does not improve readability as much as it screws up indentation. I did it nonetheless, partly to see how it looks and partly because I have no string preference for either style.

laszlocsomor:
Thank you all for the review and for the comments!
I'll wait for LGTM from @thomasvl before merging (apparently I can do that too).

laszlocsomor:
Thank you!

PR #86, opened 2018-12-06
maprule: an improved version of genrule()

laszlocsomor:
maprule() is an improved version of
native.genrule(), with the following advantages:

- Maprule can process source files in parallel,
  creating separate actions for each of them.

- Maprule does not require declaring all output
  files. Instead you declare templates for the
  output files produced for each source. Therefore
  N source files and M templates produce N*M
  outputs in total.

- Maprule supports both Bash and cmd.exe syntax
  for its commands via the specialized rules
  bash_maprule and cmd_maprule.

- Maprule's cmd attribute does deliberately not
  support $(location) expression nor Make
  Variables, in order to avoid issues and
  challenges with quoting. (Particluarly in case
  of cmd.exe passing empty arguments is
  impossible.) These paths can be passed as
  envvars instead.

- Maprule's add_env attribute does support
  $(location) expressions (and some extra
  placeholders) and is the idiomatic way to pass
  execpaths of labels in "tools" or "srcs" (the
  shared sources available for all actions) to the
  command.

See https://github.com/bazelbuild/bazel/issues/4319


laszlocsomor:
Dear reviewers,

I realize this is a huge PR. I have not found a meaningful way to split it up, or to implement the rule in pieces. Please advise if you have any suggestion to ease the review.

## Questions

1.  Does the rule's documentation (FAQ and examples in `maprule.bzl`) clarify everything?

    I have an old design doc for this rule (from 2014) that is currently not open source. It's
    outdated because Starlark has changed in the years since, and the semantics of the new `maprule`
    are somwhat different, only basic ideas are the same such as the concept of "foreach_srcs" and
    of "outs_templates", and the use of per-file actions. Notably the original design did not
    support "add_env", nor different interpreters.

2.  Where to add integration tests?

    That would mean creating actual `cmd_maprule` and `bash_maprule` targets to a `BUILD` file and
    asserting the contents of their output files (either in a `genrule` or in a custom Starlark test
    rule).

rules/maprule.bzl
```diff
@@ -0,0 +1,536 @@
...
+  They are the same except for the interpreter they use (cmd.exe and Bash respectively) and thus
+  the expected language of their "cmd" attribute. We will refer to them collectively as "maprule".
+
+- The maprule_testing struct. This should only be used by maprule's own unittests.
+
+Below you will find an Example, Tips and Tricks, and an FAQ.
+
+See the _ATTRS variable in this file for attribute documentation.
+
+## Example
```
thomasvl: Drive by: Should the example (and some of this big string, go in the `doc` on the `rule()` itself so skydoc will find it and include it when generating docs?

laszlocsomor:
Here's the `stardoc()` output: https://github.com/laszlocsomor/projects/blob/0e0c9ed15cbe697ceb2c14d0de6d9175622b39d7/bazel/maprule-doc/maprule-stardoc.md

laszlocsomor:
PTAL

rules/maprule.bzl
thomasvl: Actually, all I meant was fix how you documented things, I believe @c-parsons was working on getting stardoc going for things like this and avoiding the need for the extra copies.  So I'd suggest avoiding the duplication and just tweak your docs for now and let startdoc get wired up in an separate PR.

thomasvl: I was more the wiring in of startdoc and cloning of `dicts.add` and `paths.dirname` you mentioned.  Since Chris had seemed to indicate in a past conversation the dep cycle would be fine, it just had to be carefully rolled out, I'd was thinking avoid letting him cover that since he seemed to have a plan for it.

laszlocsomor: Sorry, I'm not sure I understood your comment -- do you suggest undoing the Stardoc wiring and copying of dicts/paths into maprule, or does @c-parsons have a plan so for now I can keep depending on dics.bzl/paths.bzl?

thomasvl: > do you suggest undoing the Stardoc wiring and copying of dicts/paths into maprule

yes.

> does @c-parsons have a plan so for now I can keep depending on dics.bzl/paths.bzl

yes, that was my understanding.

My original comment was just to move all your docs/example text off the module doc and onto the actual apis, so when stardoc does get wired (by Chris), the docs for the two apis will be more complete.

thomasvl:
fyi - deferring to others for the rule being added.

Also want to call out for @c-parsons that this adds a dependency on skydoc/stardoc and it duplicates some of dicts/paths in the new rule .bzl file due to issues with the stardoc support.  (which doesn't seem like a good thing to me)

gregestren:
Do we have any alternative APIs to Stardoc aside from reading the source?

Stardoc support sounds delightful, and could improve discoverability of all these great utilities.

thomasvl:
> Do we have any alternative APIs to Stardoc aside from reading the source?
>
> Stardoc support sounds delightful, and could improve discoverability of all these great utilities.

I don't think we need an alternative.  I was just saying that I thought @c-parsons had some idea on how to add support without having to do the duplication, so it seemed to make more sense to get get the doc strings correct until Chris gets the cycles to wire in stardoc as he was hoping.

c-parsons:
Sorry for the late responses.
I'm still a bit vacation-brained, and so can't exactly piece together a whole harmonious `bazel-skylib` -> `stardoc` dependency (cycle) off the top of my head -- I'll have to tinker with this in a separate PR.

For now, we can keep things simple by avoiding that dependency.

rules/maprule.bzl
```diff
@@ -0,0 +1,706 @@
...
+
+"""Maprule implementation in Starlark.
+
+This module exports:
+
+- The cmd_maprule() and bash_maprule() build rules.
+  They are the same except for the interpreter they use (cmd.exe and Bash respectively) and thus
+  the expected language of their `cmd` attribute. We will refer to them collectively as `maprule`.
+
+- The maprule_testing struct. This should only be used by maprule's own unittests.
```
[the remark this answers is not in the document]
laszlocsomor: Great suggestion! Done.

rules/maprule.bzl
```diff
@@ -0,0 +1,706 @@
...
+"""
+
+_cmd_maprule_intro = """
+Maprule that runs a Windows Command Prompt (`cmd.exe`) command.
+
+This rule is the same as `bash_maprule` except this one uses `cmd.exe` instead of Bash, therefore
```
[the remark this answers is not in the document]
laszlocsomor: Done.

rules/maprule.bzl
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Maprule implementation in Starlark.
+
+This module exports the cmd_maprule() and bash_maprule() build rules.
+
+They are the same except for the interpreter they use (cmd.exe and Bash respectively) and thus
+the expected language of their `cmd` attribute. We will refer to them collectively as `maprule`.
```
c-parsons: This is a little awkward -- "We will refer to them collectively as 'maprule'. [End of comment]. Perhaps if we want to provide some more context at the file-level in this aliasing file, we should "go the full mile" ?

laszlocsomor: Good point. I replaced that sentence, because the only option I saw for "going the full mile" was to copy the docstrings from maprule_private.bzl, which I didn't want to do.

How does this version look to you?

laszlocsomor:
Unfortunately I just discovered that `cmd_maprule` requires Bash because of the `ctx.resolve_command` call.

For now I won't worry about this -- I'll start a separate discussion on bazel-dev@ about my ideas to solve this.

rules/maprule.bzl
c-parsons: > @c-parsons : does your approval mean I may merge this PR?

Yes. Approval means you can merge :)

c-parsons:
So, to make sure I understand, `cmd_maprule` in this PR "works" except:

- Bash still needs to be installed
- We do `$(location)` (and other make variable) expansion unintentionally

Is that right?

laszlocsomor:
> So, to make sure I understand, `cmd_maprule` in this PR "works" except:
> 
> * Bash still needs to be installed

Yes. My proposal to rectify that: https://groups.google.com/d/msg/bazel-dev/RYlzFgzgtDc/H_peVnqjFwAJ

> * We do `$(location)` (and other make variable) expansion unintentionally

Yes, but instead of "unintentionally" I'd say "as an unnecessary side-effect of ctx.resolve_command". The rule *does* expand `$(location)` later, via `ctx.expand_location()`:
https://github.com/bazelbuild/bazel-skylib/blob/6d1c11a41f71a0d5ec9eec13f033027ce6a3804d/rules/maprule_private.bzl#L429

PR #89, opened 2019-01-02
selects.bzl: Add config_setting_group for config_setting AND/OR-chaining

gregestren:
Implements https://github.com/bazelbuild/proposals/blob/master/designs/2018-11-09-config-setting-chaining.md.

I started adding tests, but I don't think the current testing infrastructure is powerful enough for this feature. @c-parsons suggests using the upcoming [improved In-Build Starlark Testing](https://docs.google.com/document/d/17P2sgC6VPmcA7CcqC2p4cfU2Kaj6dwKiQfIXhtXTzl4/edit#heading=h.81xrc4r0s8ep) as an inaugural user. That feature's destined for the next Bazel release. Progress trackable at https://github.com/bazelbuild/bazel/issues/6237.

So now I'm just seeding the review. Should merging be hard-blocked on the above considerations?


gregestren:
CC'ing @serynth  who reviewed the proposal.

allevato:
Given that this function covers functionality in Bazel that I'm not intimately familiar with and @c-parsons has already discussed parts of the design with you (and because he's the owner of Skylib now), I prefer to defer to him on the review.

gregestren:
> and because he's the owner of Skylib now

Oh, my bad.  Hello new owner!

gregestren:
@c-parsons - given the state of https://github.com/bazelbuild/bazel/issues/6237, do you think I can start integrating this with that now?

c-parsons:
If by "integrating this with that" you mean: testing your Pull Request using the analysis-testing framework proposed in that issue?

If so, yes! Please do!
There are analysis tests using this framework which exist in the repository now today.  See https://github.com/bazelbuild/bazel-skylib/blob/master/tests/unittest_tests.bzl for some examples (in truth, they are testing the analysis-test framework itself, but these are good canonical examples to get you started)

steeve:
Is there any news preventing this to be merged ? Thanks!

gregestren:
See the previous two comments. @c-parsons was writing some Starlark testing infrastructure that this change needed to get proper testing. That wasn't ready at first but now it is. So that's what I need to integrate.

I want to tackle this in O(weeks) but it's not something I'll practically be able to have done tomorrow.

steeve:
No worries, I'm running off the patch from the PR. Thanks for the clarification!

gregestren:
Update: I'm revisiting this PR.

@c-parsons - I added a bunch of tests. It was fun using the new analysis-aware infrastructure. I would have liked to hide each test behind a single call, but that doesn't seem possible. It would also be nice to have test suite support.

I have two commented out tests which require checking that a macro `fail()`s. Is there any way to test that?

Otherwise, PTAL.

steeve:
@gregestren FYI, i've had some issues when ANDing with `config_setting` that had `constraint_values` where bazel would fail. It seems it had to do with the alias.

gregestren:
Happy to test that. Can you share an example?

steeve:
I seems it has to do with AND'ing on a platform and a config_setting:
```python
    native.config_setting(
        name = "something",
        define_values = {
            "something": "1",
        },
        visibility = ["//visibility:public"],
    )
    selects.config_setting_group(
        name = "mymegarule",
        match_all = [
            "@io_bazel_rules_go//go/toolchain:android_arm64",
            ":something",
        ],
    )
```

It fails with:
```
//:mymegarule is not a valid configuration key for //XXXX
```

steeve:
Oddly enough, if I invert the match_all, then it works
```
    selects.config_setting_group(
        name = "mymegarule",
        match_all = [
            ":something",
            "@io_bazel_rules_go//go/toolchain:android_arm64",
        ],
    )
```

steeve:
All tests on bazel 0.25.3, for the record

lib/selects.bzl
```diff
@@ -76,7 +76,198 @@ def _with_or_dict(input_dict):
...
+        # If the only entry is "//conditions:default", the condition is
+        # automatically true.
+        _config_setting_always_true(name)
+    elif not empty1:
+        _config_setting_or_group(name, match_any)
+    else:
+        _config_setting_and_group(name, match_all)
+
+def _check_duplicates(settings):
+    """ Fails if any entry in settings appears more than once.
```
c-parsons: nit: Any reason we can't just have this be a single line? (And remove the extra space at the beginning of string?

(same for other docstrings of private functions in this file)

lib/selects.bzl
```diff
@@ -76,7 +76,198 @@ def _with_or_dict(input_dict):
...
+        if settings != "//conditions:default":
+            new_settings.append(setting)
+    return new_settings
+
+def _config_setting_or_group(name, settings):
+    """ ORs multiple config_settings together (inclusively).
+
+    The core idea is to create a sequential chain of alias targets where each is
+    select-resolved as follows: If alias n matches config_setting n, the chain
+    is true so it resolves to config_setting n. Else  it resolves to alias n+1
```
c-parsons: "Else  it" -> "Else it"

gregestren:
> Oddly enough, if I invert the match_all, then it works
> 
> ```
>     selects.config_setting_group(
>         name = "mymegarule",
>         match_all = [
>             ":something",
>             "@io_bazel_rules_go//go/toolchain:android_arm64",
>         ],
>     )
> ```

So it's a TODO to be able to put `constraint_value` directly in `select()` (or by extension in `config_setting_group`). See https://github.com/bazelbuild/bazel/issues/4167#issuecomment-484237362.  So that's a valid error: all entries in `config_setting_group` have to be `config_setting`.

You can fix that in the interim by putting your `constraint_value` in a `config_setting`:

```
config_setting(
    name = "android_arm64",
    constraint_values = ["@io_bazel_rules_go//go/toolchain:android_arm64"]
)
```
As for inverting the order, an `AND` fails fast if the first condition already fails. :)

gregestren:
> 
> What are you looking for in test suite support? It looks like you somewhat do this by having all targets created under
> Is it that you'd like to be able to test specifically your whole "suite" given a single target?
> If so, I'll give it some thought :) Thanks for the feedback!

Yes, exactly. I just see a huge explosion of the number of tests at the top level so I imagine at some point of scale that could be a problem. No major issue as things stand now though.

lib/selects.bzl
gregestren: Done.

lib/selects.bzl
gregestren: Done.

lib/selects.bzl
```diff
@@ -76,7 +76,171 @@ def _with_or_dict(input_dict):
...
+    """Returns a config_setting with the given name that's always true.
+
+    This is achieved by constructing a two-entry OR chain where each
+    config_setting takes opposite values of a boolean flag.
+    """
+    name_on = name + "_stamp_binary_on_check"
+    name_off = name + "_stamp_binary_off_check"
+    native.config_setting(
+        name = name_on,
+        values = {"stamp": True},
```
HackAttack: When I try to manually create such a rule, Bazel complains
```
expected value of type 'string' for dict value element, but got True (bool)
```
so it seems like this is not valid. Is it possible this function is not being exercised?

gregestren: You're absolutely right. That's a bug and I didn't add coverage for `//conditions:default`.

PR #108, opened 2019-02-05
Add basic shell testing for unittest.bzl

c-parsons:


tests/unittest_test.sh
```diff
@@ -0,0 +1,92 @@
...
+else
+  echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
+  exit 1
+fi
+# --- end runfiles.bash initialization ---
+
+source "$(rlocation bazel_skylib/tests/unittest.bash)" \
+  || { echo "Could not source bazel_skylib/tests/unittest.bash" >&2; exit 1; }
+
+# Integration tests for bundling simple iOS applications.
```
allevato: 🤔

c-parsons: shamelessly forked from the wrong place, you got me. now I'm shamelessly forking from somewhere else (rules_apple)

c-parsons:
..and now it seems that CI only works on mac, and the error on linux is cryptic (FWIW, I don't expect a shell test to work on windows, so I'll need to disable it for that)

No idea off the top of my head what's going on, I'll need to investigate...

c-parsons:
Hello reviewers,

PTAL. Tests are now passing on ubuntu because the e2e test was tagged with "local" (by suggestion of @philwo until there's a good sandboxed bazelisk solution)

tests/unittest.bash
```diff
@@ -0,0 +1,801 @@
+#!/bin/bash
```
thomasvl: Should this be in bazel's distro given how many copies of it now seem to exist?

c-parsons: Probably, but I don't have the cycles to make this happen right now.

PR #110, opened 2019-02-14
Add analysis-test test framework

c-parsons:
This framework allows for easy creation of unittest-like tests to make assertions
on the provider-values returned by real targets.

Extensive documentation (in the form of updating the repository README) to follow in a future PR

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -72,6 +72,7 @@ def _impl_function_name(impl):
     """Derives the name of the given rule implementation function.
 
     This can be used for better test feedback."""
+
     # Starlark currently stringifies a function as "<function NAME>", so we use
     # that knowledge to parse the "NAME" portion out. If this behavior ever
     # changes, we'll need to update this.
@@ -175,7 +176,7 @@ def _make_analysis_test(impl, expect_failure = False, config_settings = {}):
 
     if changed_settings:
         test_transition = analysis_test_transition(
-            settings = changed_settings
+            settings = changed_settings,
         )
         attrs["target_under_test"] = attr.label(cfg = test_transition, mandatory = True)
     else:
@@ -277,7 +278,8 @@ def _end(env):
     if getattr(env.ctx.attr, _IS_ANALYSIS_TEST_ATTR_NAME, False):
         return [AnalysisTestResultInfo(
             success = (len(env.failures) == 0),
-            message = "\n".join(env.failures))]
+            message = "\n".join(env.failures),
+        )]
     else:
         tc = env.ctx.toolchains[TOOLCHAIN_TYPE].unittest_toolchain_info
         testbin = env.ctx.actions.declare_file(env.ctx.label.name + tc.file_ext)
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -407,7 +407,7 @@ def _expect_failure(env, expected_failure_msg = ""):
             actual_errors += cause.message + "\n"
         if actual_errors.find(expected_failure_msg) < 0:
             expectation_msg = "Expected errors to contain '%s' but did not. " % expected_failure_msg
-            expectation_msg += "Actual error:%s" % actual_errors
+            expectation_msg += "Actual errors:%s" % actual_errors
             _fail(env, expectation_msg)
     else:
         _fail(env, "Expected failure of target_under_test, but found success")
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -32,15 +32,6 @@ def register_unittest_toolchains():
 
 TOOLCHAIN_TYPE = "@bazel_skylib//toolchains/unittest:toolchain_type"
 
-# A private attribute used to denote a test rule is an analysis test.
-# For backwards compatibility, "unit tests" are not analysis tests, as being
-# an analysis test enforces the following test restrictions:
-#
-# 1. Test implementation functions may not register their own adhoc actions.
-# 2. There is a hard limit on the number of allowed transitive dependencies of
-#    analysis-test targets.
-_IS_ANALYSIS_TEST_ATTR_NAME = "_is_analysis_test"
-
 _UnittestToolchainInfo = provider(
     doc = "Execution platform information for rules in the bazel_skylib repository.",
     fields = ["file_ext", "success_templ", "failure_templ", "join_on"],
@@ -71,7 +62,8 @@ unittest_toolchain = rule(
 def _impl_function_name(impl):
     """Derives the name of the given rule implementation function.
 
-    This can be used for better test feedback."""
+    This can be used for better test feedback.
+    """
 
     # Starlark currently stringifies a function as "<function NAME>", so we use
     # that knowledge to parse the "NAME" portion out. If this behavior ever
@@ -182,8 +174,6 @@ def _make_analysis_test(impl, expect_failure = False, config_settings = {}):
     else:
         attrs["target_under_test"] = attr.label(mandatory = True)
 
-    attrs[_IS_ANALYSIS_TEST_ATTR_NAME] = attr.bool(default = True)
-
     return rule(
         impl,
         attrs = attrs,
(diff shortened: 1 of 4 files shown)


c-parsons:
Hello reviewers,

I propose deferral of writing a more-extensive README for the new framework until a future PR. I wanted to ensure the new API looked good before I went ahead and wrote a full document.
Please let me know if you'd prefer I not wait.

lib/unittest.bzl
```diff
@@ -32,6 +32,15 @@ def register_unittest_toolchains():
...
 TOOLCHAIN_TYPE = "@bazel_skylib//toolchains/unittest:toolchain_type"
 
+# A private attribute used to denote a test rule is an analysis test.
+# For backwards compatibility, "unit tests" are not analysis tests, as being
+# an analysis test enforces the following test restrictions:
+#
+# 1. Test implementation functions may not register their own adhoc actions.
+# 2. There is a hard limit on the number of allowed transitive dependencies of
+#    analysis-test targets.
+_IS_ANALYSIS_TEST_ATTR_NAME = "_is_analysis_test"
```
thomasvl: Is this some constant that also exists in bazel and has to match it?

lib/unittest.bzl
```diff
@@ -59,6 +68,19 @@ unittest_toolchain = rule(
     },
 )
 
+def _impl_function_name(impl):
+    """Derives the name of the given rule implementation function.
+
+    This can be used for better test feedback."""
```
thomasvl: Normally the closing '"""' goes on the next line.

lib/unittest.bzl
```diff
@@ -196,22 +273,27 @@ def _end(env):
...
-    tc = env.ctx.toolchains[TOOLCHAIN_TYPE].unittest_toolchain_info
-    testbin = env.ctx.actions.declare_file(env.ctx.label.name + tc.file_ext)
-    if env.failures:
-        cmd = tc.failure_templ % tc.join_on.join(env.failures)
+    if getattr(env.ctx.attr, _IS_ANALYSIS_TEST_ATTR_NAME, False):
+        return [AnalysisTestResultInfo(
+            success = (len(env.failures) == 0),
+            message = "\n".join(env.failures),
+        )]
     else:
```
thomasvl: Given the above returns, maybe drop the `else:` and drop the indent on the rest of this?

lib/unittest.bzl
```diff
@@ -308,7 +390,22 @@ def _assert_new_set_equals(env, expected, actual, msg = None):
             full_msg = expectation_msg
         _fail(env, full_msg)
 
+def _expect_failure(env, expected_failure_msg = ""):
```
thomasvl: Missing doc string since it is a public method?

lib/unittest.bzl
```diff
@@ -196,22 +273,27 @@ def _end(env):
     that the results are reported.
 
     Args:
-      env: The test environment returned by `unittest.begin`.
+      env: The test environment returned by `unittest.begin` or `analysistest.begin`.
```
thomasvl: Give the impl here forks with no overlap between analysis and unit, why both with the private attribute and conditional?  i.e. - have _end_unittest & _end_analysis, and set them into the structs independently.  You could still use the private attribute to ensure the correct begin/end are being called. As done, it actually allows the wrong begin/end to be called, if you do want to support that, then I'd suggest documenting that as an explicit impl detail.

tests/unittest_tests.bzl
```diff
@@ -10,12 +26,146 @@ def _basic_passing_test(ctx):
...
+    change_setting_fake_rule(
+        name = "change_setting_fake_target",
+        tags = ["manual"],
+    )
+
+    failure_testing_test(
+        name = "failure_testing_test",
+        target_under_test = ":failure_testing_fake_target",
+    )
+    failure_testing_fake_rule(
```
thomasvl: add blank line for consistency?

lib/unittest.bzl
c-parsons: It is not. I just wanted to pull a constant out within this file and explain its significance

tests/unittest_tests.bzl
c-parsons: The intention was to separate them into test-units (the test and the target it requires). Do you think it's worse this way?

lib/unittest.bzl
c-parsons: Done

lib/unittest.bzl
c-parsons: You're absolutely right. There was more initial overlap, and then I cleaned it up, and then the union no longer makes sense.

lib/unittest.bzl
c-parsons: Done

lib/unittest.bzl
c-parsons: (But note now this constant is no longer needed)

tests/unittest_tests.bzl
thomasvl: Ahh missed that.

lib/unittest.bzl
```diff
@@ -308,7 +389,31 @@ def _assert_new_set_equals(env, expected, actual, msg = None):
...
+
+    Args:
+      env: The test environment returned by `unittest.begin`.
+      expected_failure_msg: The error message to expect as a result of analysis failures.
+    """
+    dep = getattr(env.ctx.attr, "target_under_test")[0]
+    if AnalysisFailureInfo in dep:
+        dep_failure = dep[AnalysisFailureInfo]
+        actual_errors = ""
+        for cause in dep[AnalysisFailureInfo].causes.to_list():
```
dkelmer: And the award for most helpful PR comment goes to.....

There are fancy ways for doing this. But having said that, I'm not sure if they're more efficient/idiomatic. Nor what they are. (but I think it uses join!).

lib/unittest.bzl
```diff
@@ -308,7 +389,31 @@ def _assert_new_set_equals(env, expected, actual, msg = None):
...
+            actual_errors += cause.message + "\n"
+        if actual_errors.find(expected_failure_msg) < 0:
+            expectation_msg = "Expected errors to contain '%s' but did not. " % expected_failure_msg
+            expectation_msg += "Actual error:%s" % actual_errors
```
dkelmer: Actual errrorS. If there was more than one they'd all get listed

tests/unittest_test.sh
```diff
@@ -94,4 +104,11 @@ function test_basic_failing_test() {
   expect_log "In test _basic_failing_test from //tests:unittest_tests.bzl: Expected \"1\", but got \"2\""
 }
 
+function test_fail_unexpected_passing_test() {
+  ! bazel test //testdir:fail_unexpected_passing_test --test_output=all --verbose_failures \
```
dkelmer: Is the exclamation mark intentional? What does it do?

tests/unittest_tests.bzl
```diff
@@ -1,5 +1,23 @@
-load("//lib:unittest.bzl", "asserts", "unittest")
+load("//lib:unittest.bzl", "analysistest", "asserts", "unittest")
 
+"""Unit tests for unittest.bzl."""
+
+###################################
+####### fail_basic_test ########
```
dkelmer:
```suggestion
####### fail_basic_test ###########
```

tests/unittest_tests.bzl
```diff
@@ -10,12 +28,146 @@ def _basic_passing_test(ctx):
...
+    _change_setting_test,
+    config_settings = {
+        "//command_line_option:minimum_os_version": "1234.5678",
+    },
+)
+
+####################################
+####### failure_testing_test #######
+####################################
+def _failure_testing_test(ctx):
```
dkelmer: What are other examples of legit failures one might test for?

lib/unittest.bzl
c-parsons: a basic loop is fine with me if you don't feel strongly otherwise ;)

tests/unittest_test.sh
c-parsons: It is! It is for testing intentional failures. exclamation mark is a negation of return code.
there are two types of tests.
1.  bazel test <target> || fail "should've passed"    (the fail only occurs if bazel test failed)
2. ! bazel test <target> || fail "should've failed" (the fail only occurs if the bazel test succeded)

tests/unittest_tests.bzl
c-parsons: Well, you can theoretically test for any sort of error (including an index out of bounds exception), but I think those would be pretty terrible practice. I think the only kind of error we want to encourage expecting is one caused by fail().

tests/unittest_tests.bzl
c-parsons: Done!

lib/unittest.bzl
c-parsons: fair point! updated

PR #111, opened 2019-02-15
analysistest API for retrieval of actions registered by target under test

c-parsons:


lib/unittest.bzl
```diff
@@ -170,9 +180,16 @@ def _make_analysis_test(impl, expect_failure = False, config_settings = {}):
         test_transition = analysis_test_transition(
             settings = changed_settings,
         )
-        attrs["target_under_test"] = attr.label(cfg = test_transition, mandatory = True)
+        attrs["target_under_test"] = attr.label(
```
thomasvl: Since this duplication is getting longer, you could probably do something like:

```
    test_transition = None
    if changed_settings:
        test_transition = analysis_test_transition(
            settings = changed_settings,
        )

    attrs["target_under_test"] = attr.label(
        aspects = [_action_retrieving_aspect],
        cfg = test_transition,
        mandatory = True,
    )
```

Not sure which ends up being better, so completely up to you.

lib/unittest.bzl
```diff
@@ -412,6 +429,31 @@ def _expect_failure(env, expected_failure_msg = ""):
...
+    return dep[_ActionInfo].actions
+
+def _target_under_test(env):
+    """Returns the target under test.
+
+    Args:
+      env: The test environment returned by `analysistest.begin`.
+    """
+    result = getattr(env.ctx.attr, "target_under_test")
+    if type(result) == type([]):
```
thomasvl: types.bzl - `types.is_list(result)`

tests/unittest_tests.bzl
```diff
@@ -132,6 +132,34 @@ change_setting_with_failure_test = analysistest.make(
...
 
+####################################
+####### inspect_actions_test #######
+####################################
+def _inspect_actions_test(ctx):
+    """Test verifying actions registered by a target."""
+    env = analysistest.begin(ctx)
+
+    actions = analysistest.target_actions(env)
+    action_output = actions[0].outputs.to_list()[0]
```
thomasvl: Ensure the list of actions only has two entries?  check the second action in some way?

(I can see these tests being used as examples also on how to test, so making them more complete seems like a good thing)

lib/unittest.bzl
```diff
@@ -119,6 +119,16 @@ def _make(impl, attrs = None):
         toolchains = [TOOLCHAIN_TYPE],
     )
 
+_ActionInfo = provider()
```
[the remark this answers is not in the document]
c-parsons: I don't really think it's important for private providers, but I'll do it anyway :)

lib/unittest.bzl
c-parsons: I prefer the way it currently is, if you have no strong preference. To be re-evaluated if the code gets more complex :)

lib/unittest.bzl
c-parsons: Nice. I should really get used to using the libraries already present in this repository ;)

tests/unittest_tests.bzl
c-parsons: There's only one registered action here, but I can certainly assert the size of the actions list.

thomasvl: > There's only one registered action here, but I can certainly assert the size of the actions list.

I'm confused, `_inspect_actions_fake_rule`?  It has a `ctx.actions.declare_file()` and `ctx.actions.run_shell()`.  I guess that means one of those doesn't end up in the actions list?

c-parsons: I don't think `ctx.actions.declare_file` registers an action....

stophin:
So I'm here to ask a question, why there's no analysis_test_transition definition in unittest.bzl?
ERROR: /root/.cache/bazel/_bazel_root/603e5b5ddfbd7f179d505d1f68cc379b/external/bazel_skylib/lib/unittest.bzl:188:27: name 'analysis_test_transition' is not defined
ERROR: error loading package '': Extension 'lib/unittest.bzl' has errors
Now I can't even compile through this source!

c-parsons:
What version of bazel are you using? You can find this with:

```
bazel info
```

IIRC, Bazel has supported `analysis_test_transition` since 0.22.0. You should update Bazel if you want to use the most recent bazel-skylib versions :)

PR #117, opened 2019-02-27
maprule: use ctx.resolve_tools

laszlocsomor:
In this PR:

- Use the Bash-less ctx.resolve_tools function to
  resolve the runfiles manifests and inputs of
  tools, instead of using ctx.resolve_command for
  the same purpose.  This requires Bazel 0.23.0.

- In the _custom_envmap function: no longer
  resolve $(location) references when
  creating the envvars from custom_env, because
  those references were already resolved in
  _resolve_locations.

- Move the resolve_tools function to a standalone
  .bzl module. This function is useful for other
  rules that I'm planning to work on.

The ctx.resolve_tools() method was added in this
PR: https://github.com/bazelbuild/bazel/pull/7139
See design doc there.


laszlocsomor:
I could, let me do that in a later PR.

PR #119, opened 2019-02-27
add documentation pages for rules/ and lib/

c-parsons:


docs/collections_doc.md
```diff
@@ -0,0 +1,108 @@
+## _after_each
```
laurentlb: Looks like the support for struct is not there. I guess we'll update later.

docs/BUILD
```diff
@@ -0,0 +1,95 @@
...
+     "stardoc")
+
+# Note that due to a bug in Bazel 0.22.0, these targets cannot be built without
+# the --experimental_remap_main_repo flag.
+stardoc(
+    name = "build_test_docs",
+    input = "//rules:build_test.bzl",
+    out = "build_test_doc_gen.md",
+    deps = ["//rules:build_test"],
+)
```
laurentlb: This is very verbose.
Should there be a macro to help?

laurentlb (approved):
lgtm

c-parsons:
Sorry, this PR is not ready for review, I prematurely mailed it out when I need to clean up a couple of other commits. I'll ping this thread when it's ready for a re-review

c-parsons:
Okay. PTAL. I've cleaned things up quite a bit.

docs/BUILD
c-parsons: I don't think it's that verbose, since the verbosity is referring to explicit targets and files. I'd rather leave this as-is for now.

docs/collections_doc.md
c-parsons: Fixed!

.bazelci/presubmit.yml
```diff
@@ -6,6 +6,8 @@ tasks:
     bazel: latest
     build_targets:
     - "..."
+    build_flags:
+    - "--incompatible_remap_main_repo"
```
laurentlb: you shouldn't need this

laurentlb (approved):
lgtm

.bazelci/presubmit.yml
c-parsons: Sadly, we do, to build stardoc targets in bazel-skylib.
See https://github.com/bazelbuild/skydoc/issues/137

I've left a comment to this effect in the BUILD file, too.

docs/BUILD
ilimtvtr-alt: @googlebot

ilimtvtr-alt: @ilimtvtr-alt @c-parsons @googlebot

PR #128, opened 2019-03-15
Use //... to mean "all packages"

katre:
Using just "..." makes it appear that the command output is too long and being truncated.


katre:
Based on conversation from @dkelmer.

PR #129, opened 2019-03-18
Windows: fix tests for native test wrapper

laszlocsomor:
All tests work with
`--incompatible_windows_native_test_wrapper`
except the ones already broken on Windows
(//tests:analysis_test_e2e_test and
//tests:unittest_e2e_test).

See https://github.com/bazelbuild/bazel/issues/6622


tests/shell_tests.bzl
```diff
@@ -83,19 +83,17 @@ def _shell_spawn_e2e_test_impl(ctx):
...
-        DefaultInfo(executable = script_file),
-    ]
+    return [DefaultInfo(files = depset([script_file]))]
 
-shell_spawn_e2e_test = rule(
-    test = True,
-    implementation = _shell_spawn_e2e_test_impl,
+shell_args_test_gen = rule(
+    implementation = _shell_args_test_gen_impl,
+    outputs = {"script": "%{name}.sh"},
```
c-parsons: This syntax is subject to deprecation soon (TM)
See https://docs.google.com/document/d/14A9HK8Jn2jErMayLEE3nrNJIxNfZWN_slFbhgtS6-aM 

Could you use the recommended alternatives? (You can use `attr.output` and set the value with a macro, or just return a file as part of the returned DefaultInfo if you don't need to reference the output as its own label)

laszlocsomor: Done.

laszlocsomor:
ping

laszlocsomor:
@laurentlb ping

tests/shell_tests.bzl
```diff
@@ -83,26 +83,29 @@ def _shell_spawn_e2e_test_impl(ctx):
         "line tab\tcharacter $foo qu\"o\"te it'\\''s foo\\bar back`echo q`uote'",
         '[[ "${output}" == "${expected}" ]]',
     ])
-    script_file = ctx.actions.declare_file("%s.sh" % (ctx.label.name))
     ctx.actions.write(
-        output = script_file,
+        output = ctx.outputs.out,
```
laurentlb: Can you check if `ctx.outputs.out` was set? If not, call `declare_file`.

laszlocsomor: I made the attribute mandatory instead.

laurentlb: What's the point of the macro `shell_args_test_gen`?

laszlocsomor: To provide a default value for `out`.
`attr.output.default` is deprecated: https://docs.bazel.build/versions/master/skylark/lib/attr.html#output

laurentlb: So if you do what I suggested (declare_file is the attribute is under), you can get rid of the macro?

If you think the default value will alwaysnbe good enough, you can actually get rid of the attribute.

laszlocsomor: Done.

PR #130, opened 2019-03-19
Accept kwargs to dicts.add()

kastiglione:
This is a proposal to accept `kwargs` in `dicts.add()`. This allows `add` to be used in the same way as the `dict` constructor, which accepts keyword args in addition to another dictionary.

```
dict(other_dict, extra = 1)
dicts.add(a_dict, b_dict, extra = 1)
```


kastiglione:
Will do, thanks. Wanted to make sure it was acceptable before adding tests.

kastiglione:
thanks :+1:

PR #133, opened 2019-03-20
maprule: hide it, not ready for public use.

laszlocsomor:
Move maprule() to a private directory, to
discourage use of it. I (@laszlocsomor) am
planning breaking changes to it.

Also move private files (rule implementations) to
a subdirectory "rules/private/", to clean up the
"rules/" directory.


laszlocsomor:
All green.

PR #136, opened 2019-03-25
diff_test: add rule and tests

laszlocsomor:
This PR adds a new rule: diff_test().

The test rule compares the contents of two files.
The 'diff_test.expect_same' attribute tells
whether the files are expected to be the same or
not.

The test succeeds if the files are the same and
expected to be so, or if the files are different
and expected to be so. Otherwise the test fails.

This rules requires no Bash on Windows, so it
works without Bash, without /usr/bin/diff, and
with the Windows-native test wrapper.

See https://github.com/bazelbuild/bazel/issues/5508
See https://github.com/bazelbuild/bazel/issues/4319


laszlocsomor:
Generated Stardoc:


## diff_test

<pre>
diff_test(<a href="#diff_test-name">name</a>, <a href="#diff_test-file1">file1</a>, <a href="#diff_test-file2">file2</a>, <a href="#diff_test-expect_same">expect_same</a>, <a href="#diff_test-kwargs">kwargs</a>)
</pre>

A test that compares the contents of two files.

The test succeeds when the files are expected to be the same (with regard to
file contents) and are in fact the same, or when the files are expected to
be different and are in fact so.


### Parameters

<table class="params-table">
  <colgroup>
    <col class="col-param" />
    <col class="col-description" />
  </colgroup>
  <tbody>
    <tr id="diff_test-name">
      <td><code>name</code></td>
      <td>
        required.
        <p>
          The name of the test rule.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file1">
      <td><code>file1</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file2</code>.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file2">
      <td><code>file2</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file1</code>.
        </p>
      </td>
    </tr>
    <tr id="diff_test-expect_same">
      <td><code>expect_same</code></td>
      <td>
        optional. default is <code>True</code>
        <p>
          Whether the files are expected to be the same or not. The
  test passes if this is True and the files are the same, or if this is
  False and the files are not the same.
        </p>
      </td>
    </tr>
    <tr id="diff_test-kwargs">
      <td><code>kwargs</code></td>
      <td>
        optional.
        <p>
          The <a href="https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests">common attributes for tests</a>.
        </p>
      </td>
    </tr>
  </tbody>
</table>

laszlocsomor:
All green now, ping @laurentlb

rules/private/diff_test_private.bzl
```diff
@@ -0,0 +1,170 @@
...
+"""Implementation of diff_test macro and underlying rule.
+
+The rule compares two binary files at build time, then writes a simple script
+that is the actual test executable. The script's exit code matches the result of
+the file comparison: zero if the files are equal and were expected to be so (or
+are different and were expected to be so), non-zero otherwise.
+"""
+
+def _impl(ctx):
+    if ctx.attr.is_windows:
```
laurentlb: It may be more readable to have two functions, one for Windows, one for Linux (to avoid a long if block).

rules/private/diff_test_private.bzl
```diff
@@ -0,0 +1,170 @@
...
+        # Contents of 'test_script' depending on diff result and expectation.
+        if ctx.attr.expect_same:
+            was_same = "'#!/bin/sh\nexit 0'"
+            was_different = "\"#!/bin/sh\necho 'Expected \'$1\' and \'$2\' to be the same' >&2\nexit 1\""
+        else:
+            was_same = "\"#!/bin/sh\necho 'Expected \'$1\' and \'$2\' to be different' >&2\nexit 1\""
+            was_different = "'#!/bin/sh\nexit 0'"
+
+        # Run the shell command to generate 'test_script'.
+        ctx.actions.run_shell(
```
laurentlb: The code is a bit surprising. You run the diff as part of the build, and you generate a dummy file (exit 0 / exit 1) that will be executed during the test.

Instead, you could write the shell script with `actions.write_file`. Then the `diff` is executed as part of the test.
(the same pattern as in https://github.com/bazelbuild/examples/blob/master/rules/test_rule/line_length.bzl)

rules/private/diff_test_private.bzl
```diff
@@ -0,0 +1,170 @@
...
+if ! test -f "$1"; then echo "File not found: $1" >&2 ; exit 1; fi
+if ! test -f "$2"; then echo "File not found: $2" >&2 ; exit 1; fi
+if diff -q "$1" "$2" >/dev/null 2>&1; then echo -e {was_same} > "$3"; else echo -e {was_different} > "$3"; fi
+chmod +x "$3"
+"""
+
+        # Contents of 'test_script' depending on diff result and expectation.
+        if ctx.attr.expect_same:
+            was_same = "'#!/bin/sh\nexit 0'"
+            was_different = "\"#!/bin/sh\necho 'Expected \'$1\' and \'$2\' to be the same' >&2\nexit 1\""
```
laurentlb: it can be useful for the user to print the diff

rules/private/diff_test_private.bzl
```diff
@@ -0,0 +1,170 @@
...
+        executable = test_script,
+        files = depset(direct = [test_script]),
+        runfiles = ctx.runfiles(files = [test_script]),
+    )
+
+_diff_test = rule(
+    attrs = {
+        "file1": attr.label(mandatory = True, allow_single_file = True),
+        "file2": attr.label(mandatory = True, allow_single_file = True),
+        "expect_same": attr.bool(default = True),
```
laurentlb: Do you actually need the `expect_same` attribute?
I don't see a good use-case for that. I'd suggest to keep the rule even simpler and remove it.

rules/private/diff_test_private.bzl
laszlocsomor: Yes, with Bash that's easy (`diff ... | head`) but on Windows `fc.exe` is extremely chatty and there's no `head`-like utility.

rules/private/diff_test_private.bzl
laszlocsomor: Done.

laszlocsomor:
All green, PTAL.

Updated documentation:

## diff_test

<pre>
diff_test(<a href="#diff_test-name">name</a>, <a href="#diff_test-file1">file1</a>, <a href="#diff_test-file2">file2</a>, <a href="#diff_test-kwargs">kwargs</a>)
</pre>

A test that compares the contents of two files.

The rule compares the files at build time. If the files match, the rule
writes a dummy platform-specific script (.bat file on Windows, .sh on other
platforms) that just exits with 0, and which is the actual test script. If
the files do not match, the rule fails to build.

(The comparison is done at build time, not test time, because it makes the
rule easier to implement.)


### Parameters

<table class="params-table">
  <colgroup>
    <col class="col-param" />
    <col class="col-description" />
  </colgroup>
  <tbody>
    <tr id="diff_test-name">
      <td><code>name</code></td>
      <td>
        required.
        <p>
          The name of the test rule.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file1">
      <td><code>file1</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file2</code>.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file2">
      <td><code>file2</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file1</code>.
  be the same or not. The test passes if this is True and the files are
  the same, or if this is False and the files are not the same.
        </p>
      </td>
    </tr>
    <tr id="diff_test-kwargs">
      <td><code>kwargs</code></td>
      <td>
        optional.
        <p>
          The <a href="https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests">common attributes for tests</a>.
        </p>
      </td>
    </tr>
  </tbody>
</table>

rules/diff_test.bzl
```diff
@@ -0,0 +1,26 @@
...
+The rule uses a Bash command (diff) on Linux/macOS/non-Windows, and a cmd.exe
+command (fc.exe) on Windows (no Bash is required).
+"""
+
+load(
+    "//rules/private:diff_test_private.bzl",
+    _diff_test = "diff_test",
+)
+
+diff_test = _diff_test
```
laurentlb: Since `private/diff_test_private.bzl` has only one public symbol, both `diff_test_private.bzl` and `diff_test.bzl` are equivalent.

Maybe put all the content here and delete `diff_test_private.bzl`?

rules/private/diff_test_private.bzl
```diff
@@ -0,0 +1,161 @@
...
+    content = r"""
+set -euo pipefail
+if ! test -f "$1"; then echo "ERROR[$0]: File not found: $1" >&2 ; exit 1; fi
+if ! test -f "$2"; then echo "ERROR[$0]: File not found: $2" >&2 ; exit 1; fi
+if diff -q "$1" "$2" >/dev/null 2>&1; then
+  echo -e '#!/bin/sh\nexit 0' > "$3"
+  chmod +x "$3"
+else
+  echo "ERROR[$0]: Files \"$1\" and \"$2\" are different" >&2
+  exit 1
```
laurentlb: This means we get a broken build instead of a test failure.

laurentlb:
> Yes, but inputs must become runfiles. Finding them is [hard enough in a Bash script](https://github.com/bazelbuild/bazel/blob/3f791e7668ad1a9f6570b7d08b388eee284f3fab/tools/bash/runfiles/runfiles.bash#L23-L59), it's even harder in a Batch file.

It feels like it defeats the purpose of tests. Do we have a tracking bug? Or can you start a discussion on the mailing-list?

If our recommendation is to generate files that just do `exit 0`, I'd say something is wrong.

laszlocsomor:
> It feels like it defeats the purpose of tests.

Actually, that makes sense. Let me think it over and ping this thread later.

rules/diff_test.bzl
laszlocsomor: Done.

laszlocsomor:
PTAL.

Stardoc below.

---

## diff_test

<pre>
diff_test(<a href="#diff_test-name">name</a>, <a href="#diff_test-file1">file1</a>, <a href="#diff_test-file2">file2</a>, <a href="#diff_test-kwargs">kwargs</a>)
</pre>

A test that compares two files.

The test succeeds if the files' contents match.


### Parameters

<table class="params-table">
  <colgroup>
    <col class="col-param" />
    <col class="col-description" />
  </colgroup>
  <tbody>
    <tr id="diff_test-name">
      <td><code>name</code></td>
      <td>
        required.
        <p>
          The name of the test rule.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file1">
      <td><code>file1</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file2</code>.
        </p>
      </td>
    </tr>
    <tr id="diff_test-file2">
      <td><code>file2</code></td>
      <td>
        required.
        <p>
          Label of the file to compare to <code>file1</code>.
        </p>
      </td>
    </tr>
    <tr id="diff_test-kwargs">
      <td><code>kwargs</code></td>
      <td>
        optional.
        <p>
          The <a href="https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests">common attributes for tests</a>.
        </p>
      </td>
    </tr>
  </tbody>
</table>

laszlocsomor:
Hold on, I discovered a bug with `--nolegacy_external_runfiles`. Let me fix it and ping again.

laszlocsomor:
PTAL -- all fixed, ready for review.

laszlocsomor:
ping

rules/diff_test.bzl
```diff
@@ -0,0 +1,145 @@
...
+        return f.path[len(f.root.path) + 1:]  # generated file
+    else:
+        return f.path  # source file
+
+def _diff_test_impl(ctx):
+    if ctx.attr.is_windows:
+        test_bin = ctx.actions.declare_file(ctx.label.name + "-test.bat")
+        ctx.actions.write(
+            output = test_bin,
+            content = r"""@echo off
```
c-parsons: I find it a little awkward to have this much content written as a string literal in Starlark. Could we move it to a bat file and use that as an action input?

laszlocsomor: I tried, but couldn't: I created templates under `//rules/private` and added a label attribute to `_diff_test` that `select()`s from them:
```
    diff_script_template = select({
        "@bazel_tools//src/conditions:host_windows": "//rules/private:diff_test_tmpl.bat",
        "//conditions:default": "//rules/private:diff_test_tmpl.sh",
    }),
```

Unfortunately the rule can't find the templates when the rule itself is in an external repo.

If you don't mind, I'd leave the ugly strings here; that approach works.

I edited a test to pull the rule from an external repo.

c-parsons: I don't understand the issue you're running into ...
1. It's not necessary to select on the diff script -- you can depend on the scripts for both windows and non-windows, and only actually use the script you need based on the value of is_windows.
2. Depending on a script "tool" dependency is a pretty normal pattern for rules. If there's a serious problem with adopting this pattern from this repository, we should investigate.

That said, I know this review has been going slowly, so I'm fine with proceeding as-is and following up later. It **should** be an easy cleanup.

laszlocsomor:
Thank you @c-parsons !

laszlocsomor:
Well, this is embarrassing -- I found a simpler solution to the problem I meant to solve with diff_test.
I'm sorry for wasting your time @c-parsons , but I'd rather admit this now than carry the rule for no reason. Are you OK with me reverting this PR?

c-parsons:
Is the solution generalized to make diff_test entirely obsolete? (Care to elaborate?)
A rollback would be fine.

laszlocsomor:
No, it's specialized: https://github.com/bazelbuild/buildtools/pull/610
The solution fixes that one test macro's genrule.

diff_test is generic, but I had no purpose for it other than fixing `//api_proto:api.gen.pb.go_checkshtest` in https://github.com/bazelbuild/buildtools.

Using diff_test in Buildtools would require either to depend on Skylib as an external repo, or to copy diff_test.bzl into Buildtools.

Buildtools already depends transitively on an older Skylib version, updating the dependencies might break the affected projects. And copying diff_test into Buildtools repo risks diverging from the version in Skylib.

All in all, fixing Buildtools without diff_test is easier, but now I have no users for diff_test.

It's a nice-to-have rule, but we shouldn't maintain it until we actually need it. If I delete it now, we can revive it's required later.

WDYT?

c-parsons:
diff_test fits a pattern which is used in many places... For one, I was hoping to use it for Stardoc's tests instead of a [home-grown diff test shell script](https://github.com/bazelbuild/skydoc/blob/master/test/diff_test_runner.sh)

Hopefully that would be sufficient to justify the maintenance of diff_test ?

As a side note, I would encourage you to upgrade other repositories to the newest version of skylib :)

laszlocsomor:
Yes, that's enough reason to keep the test. Yay!

PR #138, opened 2019-04-02
Reformat with buildifier --warnings=all

thomasvl:
Fixes some dictionaries to put things in the more common orders. Buildifer
used to default to doing this check and reformatting which is why the
//conditions:default got moved up in these in the first place.


thomasvl:
The bazelci config could also get teaks to enabled the currently optional checks, or just left on the default set, but it seemed better to undo some of the sorting it had done before that move some things to non standard positions.

PR #140, opened 2019-05-01
Pass through an attribute arguments to `analysistest.make`.

sergiocampama:
This allows `analysistest.make` rules to be parameterized.

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -80,7 +80,7 @@ def _impl_function_name(impl):
     impl_name = impl_name.partition("<function ")[-1]
     return impl_name.rpartition(">")[0]
 
-def _make(impl, attrs = None):
+def _make(impl, attrs = {}):
     """Creates a unit test rule from its implementation function.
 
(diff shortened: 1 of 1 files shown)


lib/unittest.bzl
```diff
@@ -177,7 +179,7 @@ def _make_analysis_test(impl, expect_failure = False, config_settings = {}):
       A rule definition that should be stored in a global whose name ends in
       `_test`.
     """
-    attrs = {}
+    attrs = dict(attrs) if attrs else {}
```
c-parsons: Could you just make the default value of attrs be {} instead of None, and then just say attr = dict(attrs) here for mutability?
(Wouldn't be a big deal except Stardoc thus automatically detects that attrs is expected to be a dict :))

sergiocampama: do you want me to make this change in unittest.make as well?

c-parsons: ooh, I hadn't noticed we also did that there. yes please!

sergiocampama: done!

PR #146, opened 2019-05-07
Fix a number of misc issues to allow google usage of bazel-skylib

c-parsons:
1. Missing copyright header
2. Shell test fixes to use TEST_TMPDIR to have write access to directories
3. diff_test fix to use TEST_SRCDIR

### DIFF
--- a/tests/diff_test/BUILD
+++ b/tests/diff_test/BUILD
@@ -14,6 +14,7 @@ sh_test(
         "//tests:unittest.bash",
     ],
     deps = ["@bazel_tools//tools/bash/runfiles"],
+    tags = ["local"],
 )
 
 diff_test(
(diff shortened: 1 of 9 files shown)


tests/diff_test/BUILD
```diff
@@ -14,6 +14,7 @@ sh_test(
         "//tests:unittest.bash",
     ],
     deps = ["@bazel_tools//tools/bash/runfiles"],
+    tags = ["local"],
```
thomasvl: Do you want to comment on why this is needed?

c-parsons: Because the test uses bazel, which is otherwise not a test dependency of the test.

I've added a comment.

PR #156, opened 2019-05-23
Fix repository for compatibility with --incompatible_no_support_tools_in_action_inputs

laurentlb:


rules/run_binary.bzl
```diff
@@ -44,7 +44,8 @@ def _impl(ctx):
     }
     ctx.actions.run(
         outputs = ctx.outputs.outs,
-        inputs = depset(direct = ctx.files.srcs, transitive = [tool_inputs]),
+        inputs = depset(direct = ctx.files.srcs),
```
thomasvl: Can `inputs` just be ctx.files.srcs or does it have to be wrapped in a depset?

laurentlb: Thanks, the depset was indeed useless

rules/private/copy_file_private.bzl
```diff
@@ -48,7 +48,7 @@ def copy_cmd(ctx, src, dst):
 
 def copy_bash(ctx, src, dst):
     ctx.actions.run_shell(
-        inputs = [src],
+        tools = [src],
```
laszlocsomor: Why this change?
I believe we need target-config here, which I believe `inputs` is but `tools` isn't.

laurentlb: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/116#24ea5562-55a0-41c4-b439-94a288884804

```
Found tool(s) 'bazel-out/host/bin/tests/run_binary/printargs' in inputs. A tool is an input with executable=True set. All tools should be passed using the 'tools' argument instead of 'inputs' in order to make their runfiles available to the action. This safety check will not be performed once the action is modified to take a 'tools' argument. To temporarily disable this check, set --incompatible_no_support_tools_in_action_inputs=false.
```

PR #159, opened 2019-06-03
Add some comments to unittest_test.sh

c-parsons:


c-parsons:
(This PR has the added benefit of being a nice place for us to add extra test setup hooks for required Google-internal test setup)

Issue #82, opened 2018-11-29
Deprecate lib.bzl


Capstan:
This process was started in commit 1099dd2 by @thomasvl .

1. Use a `print()` warning.
1. Wait one or two releases. ← we are here.
1. Use a `fail()` error.
1. Remove `lib.bzl`

manekinekko:
Hi, I got a quick question about this deprecation. If I got this right:

BEFORE:  `load("@bazel_skylib//:lib.bzl", "paths")`
AFTER:  `load("@bazel_skylib//lib:paths.bzl", "paths")`

Right?

Capstan:
Correct. Or if you want to be pedantic about not re-exporting:

`load("@bazel_skylib//lib:paths.bzl", _paths="paths")`

manekinekko:
So, in theory, [this code](https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_repositories.bzl#L25) shouldn't trigger the deprecation warnings? Right?

Because we are getting these warning in https://github.com/angular/angular/issues/27603

Capstan:
Correct. Unfortunately, you'll still get the warning if any starlark in your transitive closure of loads loads lib.bzl, including in the `WORKSPACE` file, IIUC.

manekinekko:
Gotcha. Thanks for the feedback.

laurentlb:
In my experience, the `print` warning was not useful. Users should deal with it when they upgrade (so it should be an error). Some people didn't, because it's a warning and maybe they didn't see it. As a result, anyone building their code gets a notification in the console. It's quite noisy and it encourages everyone to ignore the messages. For most people seeing the message, it's not actionable, because it's not their code.

So let's replace it with `fail` now.

PR #162, opened 2019-06-12
lib.bzl: replace the warning with an error

laurentlb:
#82


c-parsons:
I would like to make sure that we fix internal teams before we go forward with this change
(As is, this change will prevent syncing to our monorepo)

aiuto:
This can be guarded with an incompatible flag. Then we can import it to Google, test for breakage and go. We can also add it to the incompatible flags flip CI so that for the 1.0 release we can flip it.

laurentlb:
(obsolete)

PR #167, opened 2019-07-03
Give BuildSettingInfo's value field a description

juliexxia:


rules/common_settings.bzl
```diff
@@ -23,7 +23,7 @@ https://docs.bazel.build/versions/master/skylark/config.html#user-defined-build-
 
 BuildSettingInfo = provider(
     doc = "A singleton provider that contains the raw value of a build setting",
-    fields = ["value"]
+    fields = {"value": "raw value of build settings from command line or default"}
```
c-parsons: I don't quite understand what this means :( Could you perhaps make it a little more verbose for clarity?

juliexxia: Updated (not entirely sure I did the whole git process right). PTAL!

PR #168, opened 2019-07-11
Make config_setting_group visibility aware

kwasimensah:
There might be other attributes of config_setting you want to expose. I didn't want to do a general kwargs since this is using alias underneath and that might allow leaking attributes you don't want the macro to support.

### DIFF
--- a/lib/selects.bzl
+++ b/lib/selects.bzl
@@ -76,7 +76,7 @@ def _with_or_dict(input_dict):
             output_dict[key] = value
     return output_dict
 
-def _config_setting_group(name, match_any = [], match_all = []):
+def _config_setting_group(name, match_any = [], match_all = [], visibility = None):
     """Matches if all or any of its member `config_setting`s match.
 
     Example:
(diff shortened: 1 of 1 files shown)


lib/selects.bzl
```diff
@@ -107,6 +107,7 @@ def _config_setting_group(name, match_any = [], match_all = []):
       match_all: A list of `config_settings`. This group matches if *every*
           member in the list matches. If this is set, `match_any` must be not
           set.
+      visibility: Visibiliity of the main alias used.
```
gregestren: From a user perspective, would it be clearer to call this the visibility of the config_setting_group?

Also pedantic typo nit: double i's in the second visibility!

kwasimensah: typos and comment fixed

c-parsons:
Apologies, I'm a little unclear on how this makes config_setting_group "select aware", I understand that it repropagates visibility, which is a useful feature.

As an aside, I was in a conversation recently with some other interested parties, and we tried to tabulate which attributes we'd like to recommend are almost always propagated by macro definitions, I think we came up with: `visibility`, `tags`, and `testonly`.

kwasimensah:
Meant to say "visbility" updated the title. Can we do the other attributes in a separate pull request? This is blocking me on a migration I'm trying to make and I'd like to get this is in ASAP

PR #169, opened 2019-07-12
Fix some formatting problems.

aiuto:
I just ran buildifier over the file.

### DIFF
--- a/rules/common_settings.bzl
+++ b/rules/common_settings.bzl
@@ -24,10 +24,10 @@ https://docs.bazel.build/versions/master/skylark/config.html#user-defined-build-
 BuildSettingInfo = provider(
     doc = "A singleton provider that contains the raw value of a build setting",
     fields = {
-        "value": "The value of the build setting in the current configuration. "
-        + "This value may come from the command line or an upstream transition, "
-        + "or else it will be the build setting's default."
-    }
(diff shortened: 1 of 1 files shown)


rules/common_settings.bzl
```diff
@@ -75,24 +75,26 @@ def _string_impl(ctx):
     if len(allowed_values) == 0 or value in ctx.attr.values:
         return BuildSettingInfo(value = value)
     else:
-      fail("Error setting "+ str(ctx.label) + ": invalid value '" + value + "'. Allowed values are " + str(allowed_values))
+        fail("Error setting " + str(ctx.label) + ": invalid value '" + value + "'. Allowed values are " + str(allowed_values))
```
sergiocampama: is this longer than 100 chars? it's going to complain internally

aiuto: buildifier did it. The same bits were checked in as bazelbuild/bazel/tools/config/common_settings.bzl and it did not complain then.

aiuto: Actually, it should not complain. Vendored in code is not subject to line length restrictions. It would be impossible to import stuff if that were true.

laurentlb: There's no line length limit.

Issue #163, opened 2019-06-13
0.9.0 release


robbertvanginkel:
The last version of bazel skylib was relased on March 20th, almost 3 months ago. Since then a few new rules such as diff_test have been added, and it would be great to be able to consume those from a released version.

Are there any plans to do a new skylib release soon?

Globegitter:
Running into the same issue for rules_nodejs. See https://github.com/bazelbuild/rules_nodejs/pull/863/files#diff-48449ecf6779b035ef4cc8783bd23133R77 - would be great to get a release to depend on instead.

meisterT:
See also the thread that @aiuto started on bazel-dev: https://groups.google.com/forum/#!topic/bazel-dev/fEyp04FSxFc

aiuto:
I just cut a 0.9.0 release https://github.com/bazelbuild/bazel-skylib/releases

This  fiddles with the release notes format a bit. The WORKSPACE section should be directly cut-and-pastable into your WORKSPACE.

Please take a look.

nlopezgi:
There seems to be issues with 0.9.0 release.
See PRs updating in rules_docker and bazel-toolchains:
https://github.com/bazelbuild/rules_docker/pull/992
https://github.com/bazelbuild/bazel-toolchains/pull/630

Error message in both cases is:
```
(01:24:02) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/org_golang_x_tools/go/internal/gcimporter/BUILD.bazel:18:1: in go_tool_library rule @org_golang_x_tools//go/internal/gcimporter:go_tool_library:
Traceback (most recent call last):
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/org_golang_x_tools/go/internal/gcimporter/BUILD.bazel", line 18
		go_tool_library(name = 'go_tool_library')
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/io_bazel_rules_go/go/private/rules/library.bzl", line 42, in _go_library_impl
		go.archive(go, source)
[...]

PR #179, opened 2019-07-26
Accept list parameter in new_sets union

mjduijn:
The new_sets union does not accept a list as one of its parameters, leading to a number of build breakages when updating to Skylib 0.9.0, see also https://github.com/bazelbuild/bazel-skylib/issues/163#issuecomment-511408099.
The function [documentation](https://github.com/bazelbuild/bazel-skylib/blob/2b38b2f8bd4b8603d610cfc651fcbb299498147f/lib/new_sets.bzl#L179) states that it should accept an arbitrary number of sets or lists, so the obvious fix was to make it do so :)


lib/new_sets.bzl
```diff
@@ -22,6 +22,8 @@
 
 load(":dicts.bzl", "dicts")
 
+_list_type = type([])
```
techsy730: load "types.bzl" and use "types.is_list()"?

aiuto:
@c-parsons Can you take a look at this. I was about to make the same change. It will help rules_go and rules_docker migrate to the current skylib. Otherwise they are going to have to use old_sets.bzl until they can make the bigger change.

mjduijn:
Thanks for the feedback, I'll be back from vacation this weekend and have another look at this then

mjduijn:
I'm not sure what to do about the build failure, I'm guessing it has to do with #184.
Some help would be appreciated.

aiuto:
You need to add types as a dep of new_sets.  Try this patch

index 8ae8259c5a..779812ca32 100644
--- a/lib/BUILD
+++ b/lib/BUILD
@@ -40,7 +40,10 @@ bzl_library(
 bzl_library(
     name = "new_sets",
     srcs = ["new_sets.bzl"],
-    deps = [":dicts"],
+    deps = [
+        ":dicts",
+        ":types",
+    ],
 )
 
 bzl_library(

aiuto:
Also you can't just build the docs (which seem to be failing) because of strangeness with how repositories get named.  Try this:
        bazel build docs:* --incompatible_remap_main_repo

mjduijn:
Thanks, that fixed it

techsy730:
Maybe a `to_sets` or similar parameter would be more general here?
Would turn list elements (or maybe tuples too, or even any iterable) into sets, and leave sets as is.

Yes, that basically is what `[x if types.is_set(x) else sets.make(x) for x in things_lists_list]` does, but perhaps maybe a brief survey can be done whether this is a common enough comprehension to pull into a method?

mjduijn:
Thank you for the feedback. As I have not worked intensively with Bazel and Skylib for over a year I have no idea if the referenced issue is still relevant and how best to fix it.

If it helps the discussion I can keep this PR open but I will not be make any further changes to it.

tetromino:
Ok. Closing for now, since afaik the build breakage has been already fixed elsewhere. But we would welcome a PR that addresses https://github.com/bazelbuild/bazel-skylib/pull/179#pullrequestreview-625767179 :)

PR #181, opened 2019-08-09
Add types.is_set() to test whether an arbitrary object is a set as defined by new_sets.bzl.

techsy730:
Since using sets requires special API, it can be useful to determine
whether an object is a set so special handling can be used.
For example, if a method wants to be able to take a list or a set.


techsy730:
One thing I don't know is whether "new_sets.bzl" is the right place for this method or "types.bzl" is. "types.bzl" seems to be about testing for Bazel's native types, not custom "types".

aiuto:
This looks like it should be in types.bzl not new_set.bzl

lib/new_sets.bzl
```diff
@@ -217,6 +218,19 @@ def _repr(s):
     """
     return repr(s._values.keys())
 
+_SET_TYPE = type(_make())
```
aiuto: This should be in types.bzl, alongside the other is_? predicates.

techsy730:
> This looks like it should be in types.bzl not new_set.bzl

Done

lib/new_sets.bzl
techsy730: For some reason GitHub is stating a change is requested

techsy730: Done

aiuto:
This has sat too long. If I do not hear from anyone else I will merge this on 2019-09-06.

jin:
Is this mergeable?

PR #182, opened 2019-08-09
Comply with the standards of the Bazel federation

fweikert:
This commit allows bazel_skylib to be a member of the Bazel federation, since it adds the required bzl files related to setup and dependencies.
Moreover, it also changes the WORKSPACE to fetch all dependencies through the federation.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,22 +1,30 @@
 workspace(name = "bazel_skylib")
 
-load(":workspace.bzl", "bazel_skylib_workspace")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "bazel_federation",
+    url = "https://github.com/bazelbuild/bazel-federation/archive/130c84ec6d60f31b711400e8445a8d0d4a2b5de8.zip",
+    sha256 = "9d4fdf7cc533af0b50f7dd8e58bea85df3b4454b7ae00056d7090eb98e3515cc",
+    strip_prefix = "bazel-federation-130c84ec6d60f31b711400e8445a8d0d4a2b5de8",
+    type = "zip",
+)
 
-bazel_skylib_workspace()
+load("@bazel_federation//:repositories.bzl", "bazel_skylib_deps")
+
+bazel_skylib_deps()
+
+load("@bazel_federation//setup:bazel_skylib.bzl", "bazel_skylib_setup")
+
+bazel_skylib_setup()
 
 # Below this line is for documentation generation only,
 # and should thus not be included by dependencies on
 # bazel-skylib.
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("//:internal_deps.bzl", "bazel_skylib_internal_deps")
 
-http_archive(
-    name = "io_bazel_skydoc",
-    url = "https://github.com/bazelbuild/skydoc/archive/0.3.0.tar.gz",
-    sha256 = "c2d66a0cc7e25d857e480409a8004fdf09072a1bd564d6824441ab2f96448eea",
-    strip_prefix = "skydoc-0.3.0",
-)
+bazel_skylib_internal_deps()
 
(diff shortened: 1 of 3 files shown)


thomasvl:
Don't you also need to update https://github.com/bazelbuild/bazel-skylib#workspace-file with the directions for others?

fweikert:
Yes, but we can't do that now since the federation isn't ready for general use yet. The process looks like this:

1. Bring P0 projects (skylib, cc, java, python) into compliance with federation standards [this commit].
2. Actually get a federation version that passes on CI.
3. Update documentation to tell people to use that version (or a newer one).

fweikert:
Are there any blockers on your side? I'd like to get this PR merged as soon as possible so that I can build a working federation release, which we can then advertise to users (and mention in your documentation).

WORKSPACE
```diff
@@ -1,22 +1,30 @@
 workspace(name = "bazel_skylib")
 
-load(":workspace.bzl", "bazel_skylib_workspace")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "bazel_federation",
+    url = "https://github.com/bazelbuild/bazel-federation/archive/130c84ec6d60f31b711400e8445a8d0d4a2b5de8.zip",
```
laurentlb: First line of federation still says "Experimental: do not depend on it."

Can you clarify how Skylib should use the Federation? Anything we should know?

WORKSPACE
```diff
@@ -1,22 +1,30 @@
...
+    sha256 = "9d4fdf7cc533af0b50f7dd8e58bea85df3b4454b7ae00056d7090eb98e3515cc",
+    strip_prefix = "bazel-federation-130c84ec6d60f31b711400e8445a8d0d4a2b5de8",
+    type = "zip",
+)
+
+load("@bazel_federation//:repositories.bzl", "bazel_skylib_deps")
+
+bazel_skylib_deps()
```
laurentlb: Why do we need to call this function?

WORKSPACE
```diff
@@ -1,22 +1,30 @@
...
 
-bazel_skylib_workspace()
+load("@bazel_federation//setup:bazel_skylib.bzl", "bazel_skylib_setup")
+
+bazel_skylib_setup()
```
laurentlb: Why do we need to call this function?

WORKSPACE
fweikert: We still have to clean up the federation repo before we can cut the first release. I can share an internal document with you that has more details about the project structure and how to use the federation.

WORKSPACE
fweikert: Right now the function is a no-op since skylib has no dependencies. We can delete that code for now if you want (and add it later once it's actually needed).

WORKSPACE
fweikert: bazel_skylib_setup() is basically a wrapper around bazel_skylib_workspace. It would call more functions if we added dependencies that also needed setup.

We can keep the call to bazel_skylib_workspace(), but I thought it would be nicer (and future-proof) to use the standard federation structure here.

WORKSPACE
laurentlb: What's the workflow for adding new dependencies?
First, update the function in the federation, then update skylib's dependency on the federation? Why don't we keep the list of dependencies in Skylib?

fweikert: Yes, that's the process. I know that it's cumbersome, but we wanted to have something that works - improvements to usability will be addressed later.

With the current Starlark-functions-based model I don't see how we can handle transitive dependencies when keeping the dependencies in the projects. All projects would need to define their entire dependency tree, which would lead to a lot of duplicated code across projects.
However, Tony has an idea that involves generating bzl files (if I understood him correctly).

fweikert:
How should we proceed with this PR?

c-parsons:
This LGTM. Any further comments, Laurent?

fweikert:
Thank you all!

PR #187, opened 2019-08-22
add aiuto as code owner. direwct chagnes to distribution to them

aiuto:


CODEOWNERS
```diff
@@ -1 +1,2 @@
-* @c-parsons @laurentlb @jin
+* @c-parsons @laurentlb @jin @aiuto
+distribution @aiuto @fwe
```
c-parsons: For my own information, what's the significance of "distribution changes"?

thomasvl: https://help.github.com/en/articles/about-code-owners

I'm guessing you want `distribution/*` instead.

c-parsons: Ah so this is a directory-to-be?

c-parsons: Nvm. I see the other PR.

c-parsons: Err, approved after this is changed to distribution/* :)

aiuto: Oh. duh. Yes. The empty directory shows up in my fork, so I don't notice it not being real yet.

aiuto: It doesn't need the trailing *, just the /

aiuto:
fixed the path.

PR #188, opened 2019-08-22
Add a script to regenerate the docs and update in place

aiuto:
Add a script to regenerate the docs and update in place

And run it, thus updating the docs.
Eventually these will need versioning.


aiuto:
This might fail until the remap issue is solved

docs/refresh_docs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
```
c-parsons: Could you add a documentation comment for this script?

aiuto:
Never mind. You already have a script which does this. For some reason I did not see it at all the other day.

PR #195, opened 2019-09-30
load rules thorugh maybe so we can simultaneous updates

aiuto:
We need to use maybe() for federation, so we can try updates to the federation that impact how skylib loads.  All mutually recursive repos should use maybe to make this possible.


aiuto:
When did we add this lint rule? It invalidates a lot of common practice.

PR #197, opened 2019-10-04
Avoid some repetition in _make_analysis_test

sfreilich:


lib/unittest.bzl
```diff
@@ -193,20 +193,18 @@ def _make_analysis_test(
...
-        attrs["target_under_test"] = attr.label(
-            aspects = [_action_retrieving_aspect],
-            mandatory = True,
-        )
+        target_attr_kwargs["cfg"] = test_transition
+
+    attrs["target_under_test"] = attr.label(
+        aspects = [_action_retrieving_aspect],
+        mandatory = True,
+        **target_attr_kwargs,
```
c-parsons: whoops. there's a syntax error here. See bazelCI. Maybe you cannot follow kwargs with a comma?

sfreilich: Probably. I was wondering if that's a bug since it's inconsistent with Python's behavior, but turns out it's only inconsistent with _PY3_'s behavior. Python 2 allows a comma after the final argument to a function with the exception of *args or **kwargs. (Python 3 also allows multiple sequences and/or dicts to be unpacked when calling a function.)

PR #202, opened 2019-10-07
Expose target_under_test's bin and genfiles path

sfreilich:
The output directories for the target under test may differ when the target is under a config transition (config_settings is passed to analysistest.make). Since analysis tests may assert about the command-line of generated actions, and those command-lines may contain paths to output files, this is useful information to expose.

### DIFF
--- a/tests/unittest_tests.bzl
+++ b/tests/unittest_tests.bzl
@@ -177,6 +177,54 @@ inspect_actions_test = analysistest.make(
     _inspect_actions_test,
 )
 
+########################################
+####### inspect_output_dirs_test #######
+########################################
+_OutputDirInfo = provider(fields = ["bin_path", "genfiles_path"])
+
+def _inspect_output_dirs_test(ctx):
+    """Test verifying output directories used by a test."""
+    env = analysistest.begin(ctx)
+
+    # Assert that the output dirs observed by the aspect added by analysistest
+    # are the same as those observed by the rule directly, even when that's
+    # under a config transition and therefore not the same as the output
+    # dirs used by the test rule.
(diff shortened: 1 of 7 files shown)


tests/unittest_tests.bzl
```diff
@@ -177,6 +177,63 @@ inspect_actions_test = analysistest.make(
...
+inspect_output_dirs_fake_rule = rule(
+    implementation = _inspect_output_dirs_fake_rule,
+)
+
+inspect_output_dirs_test = analysistest.make(
+    _inspect_output_dirs_test,
+    # The output directories differ between the test and target under test when
+    # the target under test is under a config transition.
+    config_settings = {
+        "//command_line_option:compilation_mode": "fastbuild",
```
c-parsons: Does this mean that these tests will now fail if tested under fastbuild?
That is, if I invoke:

```
bazel build tests:all -c=fastbuild
```

I will see test failures?

We should try to avoid that, and select at least an example option that is unlikely to be set.
How about instead:

```
"//command_line_option:test_arg": ["constant_nobody_will_use"],
```

sfreilich: That doesn't seem to be the case (at least not downstream). The targets under the config transition have different output dirs even if it ends up being a no-op.

c-parsons: Ah, interesting. Thanks.

laurentlb:
Can we avoid exposing `genfiles` and use only `bin`?
https://groups.google.com/forum/#!topic/bazel-discuss/I-mPJ_N3vws

c-parsons:
Good call. I'll make a PR

PR #207, opened 2019-10-09
Create 1.0.2 diffs from 1.0.1

aiuto:
Do not submit.
This just backs out the breaking changes from the last two days so we can release 1.0.2 instead of 2.0.0

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -126,13 +126,14 @@ def _make(impl, attrs = {}):
         toolchains = [TOOLCHAIN_TYPE],
     )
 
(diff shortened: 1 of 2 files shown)


c-parsons:
How is this PR going to be used?

The "rolled back changes" here were only introduced just before the 1.0.1 cut (they were not available in 1.0.0), and they were only available at HEAD for about an hour.

I'd like to understand the ramifications here going forward -- do any even-minutely breaking changes need to be held until we're definitely ready for the next major version?

If so, we might want to look into actually maintaining a release branch and a dev branch of this repository.

PR #208, opened 2019-10-16
Add a toolchain for Unix shell commands

aherrmann:
As suggested by @laszlocsomor in https://github.com/tweag/rules_haskell/pull/1117#issuecomment-542082833 this moves the new Unix toolchain defined in [rules_haskell](https://github.com/tweag/rules_haskell/pull/1117) to Bazel skylib.

This PR addresses https://github.com/bazelbuild/bazel/issues/5265 and the related discussion in [bazel-dev](https://groups.google.com/d/msg/bazel-dev/Lo_DKft4Cpc/yiENPDUaAQAJ). It defines a toolchain for common Unix shell commands. The toolchain provides
* paths to commands in a dictionary.
    E.g. `unix_toolchain.commands["grep"] == "/usr/bin/grep"`
* a list of binpaths suitable for defining `$PATH`.
    E.g. `unix_toolchain.paths == ["/usr/bin"]`
* make variables for each available command.
    E.g. `$(UNIX_GREP)` as in
    ```
      genrule(
          name = "use-grep",
          srcs = [":some-file"],
          outs = [":grep-out"],
          cmd = "$(UNIX_GREP) search $(execpath :some-file) > $(OUTS)",
          toolchains = [unix.MAKE_VARIABLES],
      )
    ```

The toolchain can be instantiated using `unix_configure`, a repository rule that will search the local environment for the shell commands using `repository_ctx.which`. The implementation in rules_haskell offered another way to instantiate the toolchain using rules_nixpkgs, this can be moved into rules_nixpkgs to make it more broadly available once the Unix toolchain is in bazel_skylib.

I've added a test-suite in `unix_tests.bzl`. This requires the instantiation of a unix toolchain within bazel_skylib, so I've extended `external_setup.bzl` accordingly.

### DIFF
--- a/lib/unix.bzl
+++ b/lib/unix.bzl
@@ -1,0 +1,326 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unix shell commands toolchain support.
+
+Defines a toolchain capturing common Unix shell commands, see `unix_toolchain`,
+and `unix_configure` to scan the local environment for shell commands. The list
+of known commands is available in `unix.commands`.
+
+"""
+
+load("@bazel_tools//tools/cpp:lib_cc_configure.bzl", "get_cpu_value")
+load(":paths.bzl", "paths")
+
+# List of Unix commands as specified by IEEE Std 1003.1-2008.
+# Extracted from https://en.wikipedia.org/wiki/List_of_Unix_commands.
+_commands = [
+    "admin",
+    "alias",
+    "ar",
+    "asa",
+    "at",
+    "awk",
+    "basename",
+    "batch",
+    "bc",
+    "bg",
+    "cc",
+    "c99",
+    "cal",
+    "cat",
+    "cd",
+    "cflow",
+    "chgrp",
+    "chmod",
+    "chown",
+    "cksum",
+    "cmp",
+    "comm",
+    "command",
+    "compress",
+    "cp",
+    "crontab",
+    "csplit",
+    "ctags",
+    "cut",
+    "cxref",
+    "date",
+    "dd",
+    "delta",
+    "df",
+    "diff",
+    "dirname",
+    "du",
+    "echo",
+    "ed",
+    "env",
+    "ex",
+    "expand",
+    "expr",
+    "false",
+    "fc",
+    "fg",
+    "file",
+    "find",
+    "fold",
+    "fort77",
+    "fuser",
+    "gencat",
+    "get",
+    "getconf",
+    "getopts",
+    "grep",
+    "hash",
+    "head",
+    "iconv",
+    "id",
+    "ipcrm",
+    "ipcs",
+    "jobs",
+    "join",
+    "kill",
+    "lex",
+    "link",
+    "ln",
+    "locale",
+    "localedef",
+    "logger",
+    "logname",
+    "lp",
+    "ls",
+    "m4",
+    "mailx",
+    "make",
+    "man",
+    "mesg",
+    "mkdir",
+    "mkfifo",
+    "more",
+    "mv",
+    "newgrp",
+    "nice",
+    "nl",
+    "nm",
+    "nohup",
+    "od",
+    "paste",
+    "patch",
+    "pathchk",
+    "pax",
+    "pr",
+    "printf",
+    "prs",
+    "ps",
+    "pwd",
+    "qalter",
+    "qdel",
+    "qhold",
+    "qmove",
+    "qmsg",
+    "qrerun",
+    "qrls",
+    "qselect",
+    "qsig",
+    "qstat",
+    "qsub",
+    "read",
+    "renice",
+    "rm",
+    "rmdel",
+    "rmdir",
+    "sact",
+    "sccs",
+    "sed",
+    "sh",
+    "sleep",
+    "sort",
+    "split",
+    "strings",
+    "strip",
+    "stty",
+    "tabs",
+    "tail",
+    "talk",
+    "tee",
+    "test",
+    "time",
+    "touch",
+    "tput",
+    "tr",
+    "true",
+    "tsort",
+    "tty",
+    "type",
+    "ulimit",
+    "umask",
+    "unalias",
+    "uname",
+    "uncompress",
+    "unexpand",
+    "unget",
+    "uniq",
+    "unlink",
+    "uucp",
+    "uudecode",
+    "uuencode",
+    "uustat",
+    "uux",
+    "val",
+    "vi",
+    "wait",
+    "wc",
+    "what",
+    "who",
+    "write",
+    "xargs",
+    "yacc",
+    "zcat",
+]
+
+TOOLCHAIN_TYPE = "@bazel_skylib//toolchains/unix:toolchain_type"
+MAKE_VARIABLES = "@bazel_skylib//toolchains/unix:make_variables"
+
+def _unix_toolchain_impl(ctx):
+    commands = {}
+    for cmd in _commands:
+        cmd_path = getattr(ctx.attr, cmd, None)
+        if not cmd_path:
+            cmd_path = None
+        commands[cmd] = cmd_path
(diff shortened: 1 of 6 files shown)


laszlocsomor:
Thanks for the PR! I can't review it today but will try doing it tomorrow.

lib/unix.bzl
```diff
@@ -0,0 +1,326 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
laszlocsomor: nit: 2019

lib/unix.bzl
```diff
@@ -0,0 +1,326 @@
...
+    toolchains = [TOOLCHAIN_TYPE],
+)
+
+def _unix_config_impl(repository_ctx):
+    cpu = get_cpu_value(repository_ctx)
+    commands = {}
+    for cmd in _commands:
+        cmd_path = repository_ctx.which(cmd)
+        if cmd_path == None and cpu == "x64_windows":
+            cmd_path = repository_ctx.which(cmd + ".exe")
```
laszlocsomor: If a genrule uses `$(UNIX_FOO) -out=$@` but `foo` is missing, then I guess the command will be just `-out=<path>`. It won't be obvious what exactly failed.

Could you use a fake path for nonexistent tools? Like `foo-does-not-exist`, so the command fails as `foo-does-not-exist -out=<path>`.

aherrmann: If `foo` is missing, then no make variable will be generated in the first place (see [here](https://github.com/bazelbuild/bazel-skylib/pull/208/files#diff-31822a9b4af4c9d5c8b2163235dbe860R239)). So, in that case this will be a build failure with Bazel complaining that `$(UNIX_FOO)` is undefined. I think that's preferable to a fake path, as those could be silently ignored and cause dormant issues, depending on how they're used.

laszlocsomor: True! Thanks for clarifying.

lib/unix.bzl
```diff
@@ -0,0 +1,326 @@
...
+""",
+    implementation = _unix_make_variables_impl,
+    toolchains = [TOOLCHAIN_TYPE],
+)
+
+def _unix_config_impl(repository_ctx):
+    cpu = get_cpu_value(repository_ctx)
+    commands = {}
+    for cmd in _commands:
+        cmd_path = repository_ctx.which(cmd)
```
aehlig: If we have a special toolchain configure rule, should we allow any form of user overrides, e.g., an environment variable with additional directories to look for, before looking into `PATH`, environment variables specifying the path to a tool, if I don't want to use the first on path, a mapping of alternative names.

The use case I have in mind is that authors often (accidentally) use non-POSIX extensions of their tools, so I often might have to use a particular version of the tool, which on my system might be called `gfind`, `gtest`, etc.

lib/unix.bzl
```diff
@@ -0,0 +1,326 @@
...
+        os = {
+            "darwin": "osx",
+            "x64_windows": "windows",
+        }.get(cpu, "linux"),
+        toolchain_type = TOOLCHAIN_TYPE,
+    ))
+
+_unix_config = repository_rule(
+    environ = ["PATH"],
+    local = True,
```
aehlig: This [repository rule](https://docs.bazel.build/versions/master/skylark/lib/globals.html#repository_rule) is a configure rule. Please add `configure=True`, so that it gets also invalidated upon `bazel sync --configure`.

aehlig:
> @aehlig , @aiuto, @c-parsons : what do you think?

I don't have any particular preference on where those rules end up. As repository rules often need to execute some shell commands as well, and a lot of rules also have to construct the occasional command, probably somewhere early in stack might be useful.

lib/unix.bzl
```diff
@@ -0,0 +1,326 @@
...
+      )
+""",
+    implementation = _unix_make_variables_impl,
+    toolchains = [TOOLCHAIN_TYPE],
+)
+
+def _unix_config_impl(repository_ctx):
+    cpu = get_cpu_value(repository_ctx)
+    commands = {}
+    for cmd in _commands:
```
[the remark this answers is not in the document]
aherrmann: I'm not sure what's the best way to measure this. I tried with a [JSON profile](https://docs.bazel.build/versions/master/skylark/performance.html#json-profile), by running the following command.
```
$ bazel clean --expunge && bazel fetch @local_unix_config//:all --experimental_generate_json_trace_profile
```
On my Linux machine there are two evaluator entries to `@local_unix_config` that add up to about 12ms, and another entry for `@bazel_skylib//toolchains/unix` at about 15ms. On my Windows machine the `@local_unix_config` entries add up to about 7ms, and `@bazel_skylib//toolchains/unix` is at about 7ms. (The Windows machine is a bit more powerful than the Linux machine.)

I don't know what `repository_ctx.which` does under the covers, so I'm not sure if that's plausible. The list of known commands contains 161 entries, for comparison I ran
```
$ time for i in `seq 161`; do hash foo$i >/dev/null 2>&1; done
```
which fluctuates between 20ms and 40ms.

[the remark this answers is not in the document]
laszlocsomor: `repository_ctx.which` is implemented in Java: https://github.com/bazelbuild/bazel/blob/0dd9ae90834bb08be116301553b88d0df8b0631e/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java#L512

I don't know what exactly the profile measures, but I think measuring the wall time is enough, so I did that:

1. Wrote a simple repository rule that runs 1000 `which` calls: for foo1, foo2, ... foo1000.
2. Ran `bazel query @x//:*` to have a hot Bazel server and fetch all repositories.
3. Changed "foo" in my repo rule to "bar" to ensure it's re-fetched, but nothing else is, and measured the wall time, which was ~1.6s
4. Measured the wall time of the same query without changing anything, to get a baseline, this was ~0.2s

The 1000 `which` calls took about 1.6 - 0.2 = 1.4s, so 161 of them should take about 115ms (on my machine). The same is about an order of magnitude faster on my Linux machine (they have the same amount of RAM and CPU cores).

lib/unix.bzl
aherrmann: That's a good point. I've implemented this as follows: Users can override individual commands by specifying `UNIX_<COMMAND>` environment variables. E.g. `UNIX_MAKE=/usr/bin/gmake`.

> e.g., an environment variable with additional directories to look for

Unfortunately, that's not so easy as `repository_ctx.which` doesn't allow to configure the search path, and there's no `repository_ctx.is_executable_file(path)`.

c-parsons:
> I'm having second thoughts about adding to Skylib though, maybe a separate repo like "rules-unix" would be better. IIUC Skylib's goal is to provide a standard library for the Starlark language. (In that sense, maybe adding rules to this repo -- which I have done in the past -- was a mistake, and we should move them to another repo. I don't know.)

I'm a bit torn, but would like to err on the side of caution, and say that this belongs elsewhere.

After giving it some thought, I'm inclined to agree with your "second thoughts". Skylib's mission statement is to provide standard libraries **for Starlark**, and to provide tools for Starlark rules authors (such as testing). While it's tempting to add libraries which we anticipate will be very commonly-used, this becomes a slippery slope: would we similarly suggest a Starlark `sh_library` or `sh_test` live in Skylib?

That said, the [rules](https://github.com/bazelbuild/bazel-skylib/tree/master/rules) directory is a bit beyond this mission already.

In a post-federation world where every incompatible change in Skylib requires a new major version of Skylib, I think we should err on the side of pushing back against new features.

I'm happy to discuss this further if anyone disagrees -- this is a tough call.

laszlocsomor:
Sorry I didn't answer yet.

I prefer creating a new "rules_sh" repository for it.
Putting the toolchain there vs. in Skylib allows users to depend on just that repo, not on the whole of Skylib. @c-parsons and everyone else, any objections to that?

c-parsons:
> I prefer creating a new "rules_sh" repository for it.
Putting the toolchain there vs. in Skylib allows users to depend on just that repo, not on the whole of Skylib. @c-parsons and everyone else, any objections to that?

Sounds good

laszlocsomor:
Thanks!

Meanwhile I found out that there's some administrative overhead to creating a new repo. I'll update this thread when I know more.

mboes:
@laszlocsomor Happy to proceed with a tweag/rules_sh repo in the meantime if that helps. As soon as the "administrative overhead" is dealt with, easy to transfer ownership. Can @aherrmann be included as a maintainer?

laszlocsomor:
@mboes : Great idea! It didn't occur to me before. Having the code under tweag/rules_sh would create the least friction for us, I think. @jin might want to add it to https://github.com/jin/awesome-bazel#toolchains

aherrmann:
I have opened https://github.com/tweag/rules_sh and adapted this PR to function as an independent rule set.

laszlocsomor:
Great, thank you!!

aherrmann:
Closing this in favor of [tweag/rules_sh](https://github.com/tweag/rules_sh).

PR #212, opened 2019-10-29
make select_tests resilient to default configuration differences

c-parsons:


gregestren:
Can you clarify what this does?

c-parsons:
Without this PR, select_tests would fail if tested with `-c opt`. This is because the previous test semantics were faulty.

Consider:

```
and_config_setting_group_matches_test = analysistest.make(
   _expect_matches,
   config_settings = _set_conditions([True, True, True]))
```

Using `_set_conditions([True, True, True])))` means that it will set three example flags to `True`. Notably, `compilation_mode` will be set to `opt`. Then, these flags are used to test and verify pieces of selects.bzl.

Now, consider `_set_conditions([False, False, False])))` . What this **should** do  is similarly set the three example flags to **other** values, so each of the conditions-under-test resolve to `False`. However, it would not do so prior to this change -- it would instead keep these flags at their default values.

Let me know if this is still unclear.

gregestren:
Ah, that makes enough sense. Thanks for the clarification.

PR #233, opened 2020-02-15
Create a helper rule for selecting a file from outputs of another rul…

irengrig:
…e or a filegroup by subpath


irengrig:
I found that this functionality will be useful, for instance, for rules_foreign_cc users: they build a library with CMake, and then further can select a particular file from the output (we put some of the files into predefined output groups, but we can not know all the cases, so it is not enough in the common case).

rules/select_file.bzl
```diff
@@ -0,0 +1,33 @@
...
+    implementation = _impl,
+    doc = "Selects a single file from the outputs of some target \
+by given relative path",
+    attrs = {
+        "srcs": attr.label(
+            allow_files = True,
+            doc = "The target producing the file \
+among other outputs",
+        ),
+        "subpath": attr.string(doc = "Relative path to the file"),
```
jin: `mandatory = True` and you can save a conditional in line 8.

rules/select_file.bzl
```diff
@@ -0,0 +1,33 @@
...
+    if not out:
+        fail("Can not find specified file in '%s'" % str(ctx.attr.srcs))
+    return [DefaultInfo(files = depset([out]))]
+
+select_file = rule(
+    implementation = _impl,
+    doc = "Selects a single file from the outputs of some target \
+by given relative path",
+    attrs = {
+        "srcs": attr.label(
```
jin: `mandatory = True`

tests/select_file/BUILD
```diff
@@ -0,0 +1,34 @@
...
+    subpath = "select_me.txt",
+)
+
+select_file(
+    name = "select_inner",
+    srcs = ":fg",
+    subpath = "subdir/inner.txt",
+)
+
+diff_test(
```
jin: add some unique content to the two test files? a `diff_test` between `subdir/inner.txt` and `select_me.txt` will pass, since they're both empty.

rules/select_file.bzl
```diff
@@ -0,0 +1,33 @@
...
+
+Selects a single file from the outputs of some target by given relative path.
+"""
+
+def _impl(ctx):
+    if ctx.attr.subpath and len(ctx.attr.subpath) == 0:
+        fail("Subpath can not be empty.")
+
+    out = None
+    canonical = ctx.attr.subpath.replace("\\", "/")
```
jin: do you know if we have common functionalities to deal with OS-specific path issues like this?

rules/select_file.bzl
```diff
@@ -0,0 +1,33 @@
...
+    for file_ in ctx.attr.srcs.files.to_list():
+        if file_.path.replace("\\", "/").endswith(canonical):
+            out = file_
+            break
+    if not out:
+        fail("Can not find specified file in '%s'" % str(ctx.attr.srcs))
```
jin: this would print out a wall of text of `str(ctx.attr.srcs)` is a big list, right? perhaps print it with `"\n".join`?

rules/select_file.bzl
irengrig: Not that I know of.

rules/select_file.bzl
irengrig: Done some prettifying work:)

irengrig:
Thank you for the review! PTAL

rules/select_file.bzl
```diff
@@ -0,0 +1,36 @@
...
+
+    out = None
+    canonical = ctx.attr.subpath.replace("\\", "/")
+    for file_ in ctx.attr.srcs.files.to_list():
+        if file_.path.replace("\\", "/").endswith(canonical):
+            out = file_
+            break
+    if not out:
+        files_str = ",\n".join([str(f.path) \
+for f in ctx.attr.srcs.files.to_list()])
```
jin: indentation?

irengrig: I was actually thinking that I can not use indentation if I used line continuation.
Apparently, I can :)
Corrected.

irengrig:
Hi:) We spent time reviewing and improving this - let's merge it? :)

PR #237, opened 2020-03-10
More methods for modifying dicts

CharlesG-Branch:
Just some more methods to help when working with dicts. I didn't add a test for the `unique_add` fail case because the tests in unittest_tests.bzl with `asserts.expect_failure` were marked as manual & I wanted to confirm how it should be added first.


tetromino:
Unfortunately, we cannot proceed without a signed CLA.

PR #239, opened 2020-03-24
Fix type parsing errors on "always true" conditions.

gregestren:
Specifically:

```
selects.config_setting_group(
    name = "always_true",
    match_any = ["//conditions:default"],
)
```

and

```
selects.config_setting_group(
    name = "always_true",
    match_all = ["//conditions:default"],
)
```

These should, as expected, always evaluate to True.
Their implementation had a bug that failed the build outright.

### DIFF
--- a/lib/selects.bzl
+++ b/lib/selects.bzl
@@ -234,11 +234,11 @@ def _config_setting_always_true(name, visibility):
     name_off = name + "_stamp_binary_off_check"
     native.config_setting(
         name = name_on,
-        values = {"stamp": True},
+        values = {"stamp": "1"},
     )
     native.config_setting(
         name = name_off,
-        values = {"stamp": False},
+        values = {"stamp": "0"},
     )
     return _config_setting_or_group(name, [":" + name_on, ":" + name_off], visibility)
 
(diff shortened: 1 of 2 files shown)


lib/selects.bzl
```diff
@@ -234,11 +234,11 @@ def _config_setting_always_true(name, visibility):
     name_off = name + "_stamp_binary_off_check"
     native.config_setting(
         name = name_on,
-        values = {"stamp": True},
+        values = {"stamp": "1"},
     )
     native.config_setting(
         name = name_off,
-        values = {"stamp": False},
+        values = {"stamp": "0"},
```
aiuto: I won't argue about the correctness of this, but it does seem incredibly weird that True and False don't combine to cover all values, but the string values "0" and "1" do.

Maybe I'm just grumbly. But IMO, converting ints to bools is OK from historical perspective, but strings should not.

gregestren:
@aiuto I think I need you to merge this ("Only those with write access can merge")

aiuto:
Done. I keep tripping myself up on the 3 different review/commit workflows.

PR #242, opened 2020-03-30
Stop given a depset as the first unnamed argument to the depset() constructor.

Yexo:
This is in preparation for flipping the Bazel flag --incompatible_disable_depset_items to true, making it an error to use the "items" parameter. See https://github.com/bazelbuild/bazel/issues/9017 for details.

### DIFF
--- a/bzl_library.bzl
+++ b/bzl_library.bzl
@@ -24,8 +24,11 @@ StarlarkLibraryInfo = provider(
 )
 
 def _bzl_library_impl(ctx):
-    deps_files = [x.files for x in ctx.attr.deps]
-    all_files = depset(ctx.files.srcs, order = "postorder", transitive = deps_files)
+    all_files = depset(
+        ctx.files.srcs,
+        order = "postorder",
+        transitive = [x.files for x in ctx.attr.deps],
+    )
     return [
         # All dependent files should be listed in both `files` and in `runfiles`;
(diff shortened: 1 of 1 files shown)


bzl_library.bzl
```diff
@@ -24,8 +24,11 @@ StarlarkLibraryInfo = provider(
 )
 
 def _bzl_library_impl(ctx):
-    deps_files = [x.files for x in ctx.attr.deps]
-    all_files = depset(ctx.files.srcs, order = "postorder", transitive = deps_files)
+    all_files = depset(
+        ctx.files.srcs,
```
aiuto: Don't you want direct=ctx.files.srcs?

Yexo: That's not required, the first unnamed parameter is equal to "direct".

thomasvl:
Drive by - Is the diff here just reformatting the lines and inlining the completion for transitive?  i.e. - how is this changing what goes in for the first argument?

Yexo:
> Drive by - Is the diff here just reformatting the lines and inlining the completion for transitive? i.e. - how is this changing what goes in for the first argument?

Correct, thanks for noticing. I was working on the google-internal version of the file, which didn't have https://github.com/bazelbuild/bazel-skylib/commit/1c5ed0706a8e2f73116e128184be64a99e66e83d yet. This change is basically that+formatting.

PR #243, opened 2020-04-01
Update visibility of files

laurentlb:
Stardoc needs access to a file (with exports_files) in order to
generate its documentation.


laurentlb:
cc @gregestren (this should slightly reduce the diff with Google)

PR #244, opened 2020-04-03
Depend on bzl_library, not on individual bzl files

laurentlb:


laurentlb:
cc @gregestren (this will again reduce the diff with Google)

PR #245, opened 2020-04-14
Add license and copyright notice

laurentlb:


tests/select_file/BUILD
```diff
@@ -1,6 +1,8 @@
 load("//rules:select_file.bzl", "select_file")
 load("//rules:diff_test.bzl", "diff_test")
 
+licenses(["notice"])
```
aiuto, line 4: There is  no need to bother with licenses for Bazel, it is a no-op.
Nor for tests, where people should not be depending on the code.
But this is safe either way.

laurentlb: This line is needed in Google when importing the changes. Since it is also there in other BUILD files, I think it's consistent (and simpler than trying to add a copybara transform).

Issue #248, opened 2020-05-18
feature request: Create common rule for symlinking


Yannic:
It would be great if skylib had a rule similar to `copy_file` that, instead of manually copying the file, uses `ctx.actions.symlink` (since Bazel 3.2).

This could be either implemented as a new rule (`symlink_file`?), or as attribute of `copy_file` (which could potentially become the default behavior in the future).

aiuto:
I have some reservations about this.  I don't have time to do a deep explanation this week, so I'm just asking questions about the need and the PR you sent. Because knowing the needs is important to evaluate the PR.

*why does someone need a symlink?*
- Do we just want the file under a new name and this is speed optimization over copy?
- Why do we want a new name in the first place?
- Is this for packaging later in tar or zip? (e.g. BUILD_dist in source, but need BUILD in the tgz distribution).
- Do we want to support naked symlinks, which are a different real need. Again, the motivations are the same

If we only are doing this to rename in a performant way, why can't we make this a `fast_copy` rule which promises to do a symlink if it can or copy if it can't, but it leaves a real artifiact either way.

If we are doing this to define packaging structures, then I don't think we want to build the symlinks. We want package rules that describe the need for the symlink, and the packager gets to pick what it builds.  For example, it is reasonable for a developer with remote build to have a windows host, but build a tarball that contains a symlink.

Yannic:
The purpose of this new rule is to have a faster way to give a file a new name. The reason why we need to make files available under different names is that some tools unfortunately behave differently depending on their name (e.g. `sha256(clang) == sha256(clang++)`, but they do different things when called as `clang` vs `clang++` 😡). Obviously, symlinking a large binary is faster than copying it.

I'm not sure what you mean with naked symlinks. Are you talking about the experimental unresolved symlinks in Bazel which allow creating symlinks which point to arbitrary strings (https://github.com/bazelbuild/bazel/issues/10298)? If so, then, at this moment, I don't have any intention in supporting them in the `symlink_file` rule. IMO, unresolved symlinks are tricky (caching, remote execution, ...) and should only be used to reach absolute paths outside of a WORKSPACE (e.g. `/usr/bin/gcc`). Making them easier to use will most-likely cause more harm than good.

[...]

PR #252, opened 2020-05-29
copy_file: Add parameter to allow symlinks

Yannic:
This change adds a new parameter `allow_symlinks` to `copy_file` that
allows the action to create a symlink instead of doing an expensive
copy if the execution platform (host) allows it.

Updates #248


rules/private/copy_file_private.bzl
```diff
@@ -78,9 +85,10 @@ def _ximpl(ctx):
     return _common_impl(ctx, True)
 
 _ATTRS = {
-    "src": attr.label(mandatory = True, allow_single_file = True),
-    "out": attr.output(mandatory = True),
+    "is_symlink": attr.bool(mandatory = True),
```
aiuto, line 88: It looks like the is_executable should be folded in here as another bool attribute.
Then you would not need _copy_file vs copy_xfile and the `if is_executable` at line 127.
Having two different techniques for doing this differentiation is confusing.

rules/private/copy_file_private.bzl
```diff
@@ -118,6 +129,54 @@ def copy_file(name, src, out, is_executable = False, **kwargs):
             name = name,
             src = src,
             out = out,
+            is_symlink = False,
+            is_windows = select({
```
aiuto, line 133: host_window is only true if host==target.    See the issue for concernts.

rules/private/copy_file_private.bzl
Yannic: AFAIK, we also have to set `is_executable = True` on the rule to allow it as tool for other actions.

Yannic:
Updated the PR according to the discussion in #248. ptal, thanks!

rules/private/copy_file_private.bzl
aiuto: Yes, but it would be an attribute on copy_file, and you would just pass the value down from the macro to the rule. The virtually duplicated call at lines 131-150 would fold away.

rules/private/copy_file_private.bzl
```diff
@@ -128,27 +123,20 @@ def copy_file(name, src, out, is_executable = False, allow_symlink = False, **kw
           handle symlinks (which most UNIX tools can).
       **kwargs: further keyword arguments, e.g. `visibility`
     """
+
+    copy_file_impl = _copy_file
     if is_executable:
```
aiuto, line 128: Silly github UI won't let me comment  on line 96. There is no more _ximpl.

rules/private/copy_file_private.bzl
```diff
@@ -58,12 +58,12 @@ def copy_bash(ctx, src, dst):
         use_default_shell_env = True,
     )
 
-def _common_impl(ctx, is_executable):
+def _impl(ctx):
```
aiuto, line 61: There are several rules and implementations here.
It would be easier to read if this were _copy_file_impl

rules/private/copy_file_private.bzl
Yannic: I *think* this was already fixed in 3f0948767e476c0b301caa95ab7f861a8314fcb1.

rules/private/copy_file_private.bzl
Yannic: Sure, done

Yannic:
Sorry for the delay, I've been busy doing other things!

Updated the PR to address your comments. ptal, thanks!

rules/private/copy_file_private.bzl
```diff
@@ -111,27 +114,29 @@ def copy_file(name, src, out, is_executable = False, **kwargs):
...
+
+    copy_file_impl(
+        name = name,
+        src = src,
+        out = out,
+        is_windows = select({
+            "@bazel_tools//src/conditions:host_windows": True,
+            "//conditions:default": False,
+        }),
+        is_executable = is_executable,
```
aiuto, line 139: If this were     executable = is_executable  would you just be able to call _copy_file() and not have _copy_xfile()?

Yannic: AFAICT, no. `_copy_xfile = rule(executable = True)` has a different (native) parent rule than `_copy_file = rule(executable = False)` (https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/skylark/StarlarkRuleClassFunctions.java;l=290;drc=8b856f5484f0117b2aebc302f849c2a15f273310).

Yannic:
Updating the branch did fix CI, thanks for the review!

Yannic:
Is there anything left to get this merged?

aiuto:
Nope. Other than a reminder. It turns out last week was busier than I thought and it slipped my mind.

PR #253, opened 2020-06-02
Add `new_sets` as a dep of `sets` bzl_library

Bocete:
Without this change, evaluating bzl with strict dependencies results in an error.

[`load(":new_sets.bzl", _sets = "sets")`](https://github.com/bazelbuild/bazel-skylib/blob/560d7b2359aecb066d81041cb532b82d7354561b/lib/sets.bzl#L17) results in `No such file or directory: 'third_party/bazel_skylib/lib/new_sets.bzl`


Bocete:
Any ideas on what Buildifier found wrong? All I see is a list of warnings from unaffected code, and a `Error: The command exited with status 4`.

Thanks! I'm completely new to github CI so I appreciate your patience.

Bocete:
@jin @laurentlb PTAL, this is a trivial one. Thanks!

PR #257, opened 2020-07-08
Process wrapper

jelmansouri:
As we are working toward having native rule compatibility with windows, we encountered a lot of run_shell usage to have access to some basic shell functionality like substituting $pwd in command line arguments or environment variables or redirecting the output. Or reading a file that contains arguments generated by another rule (build.rs scripts in rust).


aiuto:
Could you start with a design discussion somewhere?

I *think* I agree something like this is worthwhile because bash scripts have no place in a portable world, but there are some problems
- this takes a dependency on C++, which I'm not sure we want in skylib
- it's not clear the features implemented work in a general enough way to be useful outside of your use case.

jelmansouri:
I created a google doc to discuss this!
https://docs.google.com/document/d/13M9ArI_Qdj3PuJk2kXSRm6YISp8uceq_lEG8_I7kXcA

jin:
cc @meteorcloudy

aiuto:
+bazel-discuss <bazel-discuss@googlegroups.com>

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
...
+        </p>
+        <p>
+          executable_path: path to the executable that is going to be launched as a child process.
+        </p>
+        <p>
+          [args]: Arguments to be passed on to the child process.
+        </p>
+      </td>
+    </tr>
+    <tr id="process_wrapper-subst-pwd">
```
meteorcloudy, line 75: Should this be --subst <env var name> ? So that you can substitute any given environment variable in argument. In cmd.exe, there is no default %PWD% env var, but %CD% as the current working directory.

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
...
+      <td>
+        optional.
+        <p>
+          Writes the standard output of the child process to a file.
+          Can appear once.
+        </p>
+      </td>
+    </tr>
+    <tr id="process_wrapper-touch-file">
+      <td><code>--touch-file file_path</code></td>
```
meteorcloudy, line 121: Does this file only get created if the child process exits successfully? If so, can you make it clear in the doc or even the flag name?

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
+## process_wrapper
+
+Process wrapper is a helper that allows you, in a platform independent way,
```
meteorcloudy, line 3: I think we should make it clear that using this means you'll have C++ toolchain in your dependency.

lib/process_wrapper/process_wrapper.cc
```diff
@@ -0,0 +1,109 @@
...
+    if (arg == PW_SYS_STR("--subst-pwd")) {
+      subst_pwd = true;
+    } else {
+      // All argument in the else block have a parameter
+      if (++i == argc) {
+        std::cerr << "process wrapper error: argument \"" << ToUtf8(arg)
+                  << "\" missing parameter." << std::endl;
+        return -1;
+      }
+      if (arg == PW_SYS_STR("--env-file")) {
```
meteorcloudy, line 43: Have you tested if the process wrapper works or not when the paths given by `--env-file`, `--arg-file`, etc. are long paths on Windows (>260 characters). 
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
...
+    args.add("--current-dir", "<pwd>")
+    env = {"CURRENT_DIR": "<pwd>/test_path"}
+
+    ctx.actions.run(
+        executable = ctx.executable._process_wrapper,
+        inputs = ctx.files.env_files + ctx.files.arg_files,
+        outputs = [stdout_output],
+        arguments = [args],
+        env = env,
+        tools = [ctx.executable._process_wrapper_tester],
```
ash2k, line 28: `_process_wrapper_tester` is undefined?

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
...
+    <tr id="process_wrapper-env-file">
+      <td><code>--env-file file_path</code></td>
+      <td>
+        optional.
+        <p>
+          Loads an environment variables file form "file_path".
+          Can appear multiple times.
+        </p>
+        <p>
+          The file consists of new line separated environment variables under the form of VAR=VALUE.
```
ash2k, line 93: Please provide a way to use new line characters in values, some sort of escaping scheme. Perhaps a helper to do such escaping too.

docs/process_wrapper_doc.md
```diff
@@ -0,0 +1,130 @@
...
+    <tr id="process_wrapper-arg-file">
+      <td><code>--arg-file file_path</code></td>
+      <td>
+        optional.
+        <p>
+          Loads a command line arguments file form "file_path".
+          Can appear multiple times.
+        </p>
+        <p>
+          The file consists of  new line separated arguments.
```
ash2k, line 106: Same as above - please provide a way to use new line characters in arguments.

lib/process_wrapper/process_wrapper.cc
jelmansouri: They don't, the wrapper is going to fail, with a message like this one:
process wrapper error: failed to open file: tests/process_wrapper/test_files/lllllllllllllllllllllllllllllllllllllllllllllllooooooooooooooooooooooooooooooooooooooooooooooooonnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggggg

docs/process_wrapper_doc.md
jelmansouri: Fixed in the doc!

docs/process_wrapper_doc.md
jelmansouri: Added to the doc

docs/process_wrapper_doc.md
jelmansouri: Thanks!

lib/process_wrapper/process_wrapper.cc
meteorcloudy: Yeah. If you want the process wrapper to work for long path.
I guess you'll have to convert the paths to a format like `\\?\D:\long\windows\path`.

jmmv:
Given that Bazel already has a process-wrapper tool, should this be called something else?

jelmansouri:
> Given that Bazel already has a process-wrapper tool, should this be called something else?

I don't have any strong opinion about the name :) if you have a suggestion I would be happy to change it!

lib/process_wrapper/process_wrapper.cc
jelmansouri: Looking at the state of long path support on windows, if we are operating on paths longer than 260, it is highly probable that other tooling that is used along with the wrapper will break.

jelmansouri:
Updated the design doc with more information!

lib/process_wrapper/process_wrapper.cc
meteorcloudy: I see. Can you help add a comment about this tool not supporting long path on Windows? If other people try to use this, they should have the correct expectation.

docs/process_wrapper_doc.md
jmmv: Agreed. And even make this _not_ rely on the environment. Make this `--subst=var=value` so this can apply arbitrary substitutions.

lib/BUILD
```diff
@@ -98,3 +99,28 @@ bzl_library(
     name = "old_sets",
     srcs = ["old_sets.bzl"],
 )
+
+cc_binary(
+    name = "process_wrapper",
+    srcs = [
+        "process_wrapper/process_wrapper.cc",
```
jmmv, line 106: Any reason for not adding a `process_wrapper/BUILD` file instead?

lib/process_wrapper/process_wrapper.cc
```diff
@@ -0,0 +1,109 @@
...
+        }
+      } else if (arg == PW_SYS_STR("--arg-file")) {
+        if (!ReadFileToArray(argv[i], file_arguments)) {
+          return -1;
+        }
+      } else if (arg == PW_SYS_STR("--touch-file")) {
+        if (!touch_file.empty()) {
+          std::cerr << "process wrapper error: \"--touch-file\" appears more "
+                       "than once"
+                    << std::endl;
```
jmmv, line 55: Please avoid using `std::endl` and instead put `\n` in the strings. `std::endl` forces a flush of the stream and we don't need this here, and that's already implied by the use of stderr.

lib/process_wrapper/system_posix.cc
```diff
@@ -0,0 +1,157 @@
...
+int System::Exec(const System::StrType& executable,
+                 const System::Arguments& arguments,
+                 const System::EnvironmentBlock& environment_block,
+                 const StrType& stdout_file) {
+  StdoutPipe stdout_pipe;
+  if (!stdout_file.empty() && !stdout_pipe.CreateEnds()) {
+    return -1;
+  }
+
+  pid_t child_pid = fork();
```
jmmv, line 113: Why do we need to fork at all? From what I see in the description of this wrapper, there is no need to do anything upon process termination, so the wrapper can set things up and then give control to the wrapper process, keeping the process tree cleaner (and being much faster e.g. on macOS).

lib/BUILD
jelmansouri: I had that originally but I find it cleaner to do this @bazel_skylib//lib:process_wrapper where referencing it, rather than @bazel_skylib//lib/process_wrapper:process_wrapper.
Maybe I should have an alias in //lib ?

lib/process_wrapper/system_posix.cc
jelmansouri: The touch file feature, needs getting back the control.

lib/BUILD
jmmv: `@bazel_skylib//lib/process_wrapper:process_wrapper` is equivalent to `@bazel_skylib//lib/process_wrapper` (the default target name is the package's leaf name).

jelmansouri: I know it's the case in general, but it didn't seem to work in this context:
"_process_wrapper": attr.label(
            default = "@bazel_skylib//lib:process_wrapper",
            executable = True,
            allow_single_file = True,
            cfg = "exec",
        ),

jelmansouri:
> Also, I'm wondering... all this argument rewriting... why cannot it be done from Starlark? I think the design doc should better explain why a separate tool is necessary. A separate tool feels a bit awkward to me, and it risks becoming its own "shell interpreter" over time.

Ideally it would, but looking at some open issues dating back to 2018 like this one https://github.com/bazelbuild/bazel/issues/5511, it might be harder to make these changes to support remote executions platfoms. But I'm not an expert here!

lib/BUILD
jmmv: If that's the code snippet you tried and failed... the `:` is wrong in the `default` value. Should be a slash.

jelmansouri: Had a slash in there I think, but my memory might fail me. Let me try again!

lib/process_wrapper/process_wrapper.cc
jelmansouri: Fixed!

docs/process_wrapper_doc.md
jelmansouri: Fixed

docs/process_wrapper_doc.md
jelmansouri: Fixed

docs/process_wrapper_doc.md
jelmansouri: Updated the design document and code regarding this.

lib/BUILD
jelmansouri: Fixed

blg666 (approved):

PR #259, opened 2020-07-23
Stop depending on rules_pkg through the federation.

aiuto:
- while the federation code is in flux, this will mean less churn for skylib
- rules_pkg is only needed by developers for making a distribution, so this won't impact users.
- when we develop a new federation model, we can re-depend the right way.

fixes https://github.com/bazelbuild/bazel-federation/issues/117

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,6 +3,17 @@ workspace(name = "bazel_skylib")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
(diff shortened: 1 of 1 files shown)


aiuto:
> Should we update https://github.com/bazelbuild/bazel-skylib/blob/master/internal_deps.bzl?

I don't think so. From the point of view of skylib itself, that is currently a no-op. But if someone does include skylib via the federation and does not include python in their workspace they could conceivably have a problem. Leaving it this way should smoothly work when there is a federation update to the newer rules_pkg. The end user will be selecting the newer federation and will end up pulling in the newer rules_pkg to their workspace.

PR #260, opened 2020-07-28
rm last usage of --experimental_build_setting_api

juliexxia:


c-parsons:
Failure here looks legitimate. Looks like Stardoc still reflects this flag as being false. You have two options here:

1. Update Stardoc version (not sure when the last Stardoc release was relative to this flag change, but you may need to rebuild/release a new Stardoc. Then, update skylib's dependency on Stardoc to the new version. Then, you should be able to carry on with this commit.

OR

2. Revert this commit and keep status quo; when skylib needs to update stardoc version, we will need to also include the diff of this commit.

juliexxia:
Not sure why this is failing with the --experimental_build_setting_api check. That flag has defaulted to true for ~a year now in [bazel](https://github.com/bazelbuild/bazel/blob/3762e94856dda91f91a5d14fcd56eb412a281c68/src/main/java/com/google/devtools/build/lib/packages/StarlarkSemanticsOptions.java#L90) and it seems like the checks are running 3.4.1 so not a super old version or anything

c-parsons:
It looks like stardoc jars were updated 10 months ago (https://github.com/bazelbuild/stardoc/pull/13/commits/9346c62c8f62df22713463f8a453bb12bf08d400)

But this flag was flipped 13 months ago
(https://github.com/bazelbuild/bazel/commit/f113c4493989ddaf131fcc90020b68222f4a5bbe)

So it must just be the case that skylib is depending on a *very* old version of Stardoc, via the federation tar.
Perhaps you can update that dependency here and this should be fine?

aiuto:
Hi Julie.  I made a PR to update stardoc in the federation. If you approve
that I can make a new release of that, and you can update the skylib dep in
your PR.
https://github.com/bazelbuild/bazel-federation/pull/119

juliexxia:
@aiuto that bazel_federation update got this all fixed thanks much!

aiuto:
Great.

PR #266, opened 2020-08-18
Set config_settings in inspect_output_dirs_test to something not used…

comius:
… in the main build.

This resolves test failures, that were started by
https://github.com/bazelbuild/bazel/commit/a7a0d48fbeb059ee60e77580e5d05baeefdd5699


brandjon:
LGTM, but apparently I'm not a codeowner in this repo.

TL;DR for an owner: The linked change made no-op starlark transitions no longer spuriously change the output dir. The test in question relies on the output dir being changed. This PR uses a non-default value for the transition so it's no longer a no-op.

comius:
Closing if favour of pull/267

PR #271, opened 2020-09-03
Add support for gazelle handling relative imports

segiddins:
This is the first time I've written more than a line of go, so it's possible I'm doing things very very wrong here


achew22:
@segiddins looks good to me! Thanks so much for contributing your fix upstream. Sorry for the slow response. I was gone over the long weekend.

achew22 (approved):

segiddins:
🙏 and absolutely no need to apologize, a weekend is not a slow response at all!

PR #274, opened 2020-09-14
Handle "internal" directory visibility

achew22:
Both "internal" and "private" directories should be treated the same way
and have visibility restrictions be placed upon them.

### DIFF
--- a/gazelle/bzl/gazelle.go
+++ b/gazelle/bzl/gazelle.go
@@ -233,11 +233,10 @@ func (*bzlLibraryLang) GenerateRules(args language.GenerateArgs) language.Genera
 
 		r.SetAttr("srcs", []string{f})
 
-		if args.File == nil || !args.File.HasDefaultVisibility() {
-			inPrivateDir := pathtools.Index(args.Rel, "private") >= 0
-			if !inPrivateDir {
-				r.SetAttr("visibility", []string{"//visibility:public"})
-			}
+		shouldSetVisibility := args.File == nil || !args.File.HasDefaultVisibility()
+		if shouldSetVisibility {
+			vis := checkInternalVisibility(args.Rel, "//visibility:public")
+			r.SetAttr("visibility", []string{vis})
 		}
 
 		fullPath := filepath.Join(args.Dir, f)
@@ -334,4 +333,17 @@ func (s srcsList) Contains(m string) bool {
 		}
 	}
 	return false
+}
+
+// checkInternalVisibility overrides the given visibility if the package is
+// internal.
+func checkInternalVisibility(rel, visibility string) string {
+	if i := pathtools.Index(rel, "internal"); i > 0 {
+		visibility = fmt.Sprintf("//%s:__subpackages__", rel[:i-1])
+	} else if i := pathtools.Index(rel, "private"); i > 0 {
+		visibility = fmt.Sprintf("//%s:__subpackages__", rel[:i-1])
+	} else if pathtools.HasPrefix(rel, "internal") || pathtools.HasPrefix(rel, "private") {
+		visibility = "//:__subpackages__"
+	}
+	return visibility
 }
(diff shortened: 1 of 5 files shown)


achew22:
@jayconrod and @segiddins, could I have you both take a look at this one also. I believe this addresses the last issue that @segiddins ran into when using this in [rules_swift](https://github.com/bazelbuild/rules_swift/pull/482). If that's not the case, please help me see the things that I am missing so I can make this as easy to use as possible.

Thanks so much!

Once you two LGTM I will send off to someone from the skylib team for a last review.

gazelle/bzl/gazelle.go
```diff
@@ -324,3 +322,20 @@ func (s srcsList) Contains(m string) bool {
...
 	return false
 }
+
+// checkInternalVisibility overrides the given visibility if the package is
+// internal. This is taken from Gazelle's rule package. Unfortunately bzl
+// allows for both the name "internal" and "private" to indicate visibility and
+// so it can't be used verbatim, but with slight modifications it can be.
+// https://github.com/bazelbuild/bazel-gazelle/blob/2078a44d2bba287db98b2b48b5f6cf4f08642489/rule/rule.go#L833:6
+func checkInternalVisibility(rel, visibility string) string {
+	rel = rel + "/" // Append a / to make finding the suffix easier.
```
jayconrod, line 332: Maybe use `github.com/bazelbuild/bazel-gazelle/pathtools.Index` instead? It's like `strings.Index`, but it only matches on path component boundaries.

gazelle/bzl/gazelle.go
```diff
@@ -324,3 +322,20 @@ func (s srcsList) Contains(m string) bool {
 	}
 	return false
 }
+
+// checkInternalVisibility overrides the given visibility if the package is
+// internal. This is taken from Gazelle's rule package. Unfortunately bzl
+// allows for both the name "internal" and "private" to indicate visibility and
```
jayconrod, line 328: I don't think `rule.CheckInternalVisibility` necessarily needs to be mentioned here. Arguably that should have been in `language/go`. Each language may have its own definition of internal visibility, so I'd expect each extension might have a function like this.

jayconrod (approved):
Looks good with a couple minor comments.

segiddins (approved):

gazelle/bzl/gazelle.go
achew22: Done

gazelle/bzl/gazelle.go
achew22: I made the change and it is passing tests so I believe it is equivalent. The only reason I didn't use `pathtools.Index` instead from the start is because it wasn't looking for the `LastIndex` and I didn't want it to turn `/foo/internal/bar/internal` into being visible to `//foo:__subpackages__`

achew22:
@jayconrod with the submission of the [`CODEOWNERS`](https://github.com/bazelbuild/bazel-skylib/pull/275) I anticipate you should now have permission to approve. Could I have you give it one more poke just to make sure it is actually working now.

Thanks again!

achew22:
@jayconrod could you try approving this one more time now that we've updated the CODEOWNERS?

jayconrod (approved):
Still LGTM

jayconrod:
Actually, a couple updates are needed.

* "This branch is out-of-date with the base branch". Not sure if there are conflicts, I'd guess it just needs to be updated.
* Tests are failing at tip.
* "At least 1 approving review is required by reviewers with write access." I don't have write access, and I don't think CODEOWNERS grants it. Someone on the Bazel team needs to merge or grant us write access. I can ping someone when the other two things are done.

achew22:
@c-parsons would it be possible to add writer permissions to either (or both) me and @jayconrod

c-parsons:
@achew22 you have already had write permissions for the github repo. I've added @jayconrod now.

PR #275, opened 2020-09-14
Fix CODEOWNERS

achew22:
Previously permissions were granted to edit any file that existed in a directory
titled `gazelle`. Now grants have been given recursively.

Sorry for the inconsistency with Google3 `OWNERS` files.

@aiuto & @c-parsons . please review this very carefully since I have no prior 
experience with CODEOWNERS. I can say, however, that neither @jayconrod
nor I presently have the ability to submit code in the `/gazelle` dir, which is
why I'm messing with this.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,5 @@
+# More details on syntax here: https://help.github.com/articles/about-codeowners/
 * @c-parsons @laurentlb @jin @aiuto
-distribution/ @aiuto @fwe
-rules/ @juliexxia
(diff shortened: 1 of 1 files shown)


achew22:
I tried to send this to a subset of the root owners, but that doesn't appear to be possible. Nor does it seem possible for me to remove you. Sorry to broadcast with such a high replication factor!

achew22:
Friendly ping

c-parsons (approved):

PR #276, opened 2020-09-24
Enable unittest.suite to accept partial calls of test rules

### LINKED ISSUE #98: How does unittest support attributes on the tests?
[`unittest.make()`](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/unittest.bzl#L62) is documented at taking an optional `attrs` that will be used from then `rule` is created for the test.

[`unittest.suite()`](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/unittest.bzl#L118) is documented to create the rules from the list given to it, but in doing so, [only passes the `name` when creating the rules](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/unittest.bzl#L163-L166).

How does one set the attributes for the testing rules created?

dws-uber:
This permits using `unittest.suite` with test rules that have nondefault
attributes, while retaining compatibility with current usage.

For instance, this permits setting a `timeout` on each test in a
`unittest.suite`.  Previously, all tests in a `unittest.suite` would
have the default timeout, with no good way to alter this.  This
made it hard to eliminate all the warnings produced from using the
`--test_verbose_timeout_warnings` bazel option.

While timeouts were the motivation, the solution here is not specific
to timeouts. It will permit arbitrary additional arguments to the test
rules in a `unittest.suite`.

Fixes #98

### DIFF
--- a/lib/partial.bzl
+++ b/lib/partial.bzl
@@ -19,6 +19,11 @@ Partial function objects allow some parameters are bound before the call.
 Similar to https://docs.python.org/3/library/functools.html#functools.partial.
 """
 
+# create instance singletons to avoid unnecessary allocations
+_a_dict_type = type({})
+_a_tuple_type = type(())
+_a_struct_type = type(struct())
+
 def _call(partial, *args, **kwargs):
     """Calls a partial created using `make`.
 
@@ -124,7 +129,27 @@ def _make(func, *args, **kwargs):
     """
     return struct(function = func, args = args, kwargs = kwargs)
 
+def _is_instance(v):
+    """Returns True if v is a partial created using `make`.
+
+    Args:
+      v: The value to check.
+
+    Returns:
+      True if v was created by `make`, False otherwise.
+    """
+    # Note that in bazel 3.7.0 and earlier, type(v.function) is the same
+    # as the type of a function even if v.function is a rule. But we
+    # cannot rely on this in later bazels. Since this check is heuristic
+    # anyway, we simply check for the presence of a "function" attribute
+    # without checking its type.
+    return type(v) == _a_struct_type \
+        and hasattr(v, "function") \
+        and hasattr(v, "args") and type(v.args) == _a_tuple_type \
+        and hasattr(v, "kwargs") and type(v.kwargs) == _a_dict_type
+
 partial = struct(
     make = _make,
     call = _call,
+    is_instance = _is_instance,
 )
--- a/lib/partial.bzl
+++ b/lib/partial.bzl
@@ -140,9 +140,11 @@ def _is_instance(v):
     """
     # Note that in bazel 3.7.0 and earlier, type(v.function) is the same
     # as the type of a function even if v.function is a rule. But we
-    # cannot rely on this in later bazels. Since this check is heuristic
-    # anyway, we simply check for the presence of a "function" attribute
-    # without checking its type.
+    # cannot rely on this in later bazels due to breaking change
+    # https://github.com/bazelbuild/bazel/commit/e379ece1908aafc852f9227175dd3283312b4b82
+    #
+    # Since this check is heuristic anyway, we simply check for the
+    # presence of a "function" attribute without checking its type.
     return type(v) == _a_struct_type \
         and hasattr(v, "function") \
         and hasattr(v, "args") and type(v.args) == _a_tuple_type \
--- a/lib/types.bzl
+++ b/lib/types.bzl
@@ -138,20 +138,6 @@ def _is_set(v):
     """
     return type(v) == _a_struct_type and hasattr(v, "_values") and _is_dict(v._values)
 
-def _is_partial(v):
-    """Returns True if v is a partial function object created by partial.make().
-
-    Args:
-      v: The value whose type should be checked.
-
-    Returns:
-      True if v was created by partial.make(), False otherwise.
-    """
-    return type(v) == _a_struct_type \
-        and hasattr(v, "function") \
-        and hasattr(v, "args") and _is_tuple(v.args) \
-        and hasattr(v, "kwargs") and _is_dict(v.kwargs)
-
 types = struct(
     is_list = _is_list,
     is_string = _is_string,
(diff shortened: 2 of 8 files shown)


dws-uber:
Looks like the buildifier issues in CI are pre-existing problems: https://buildkite.com/bazel/bazel-skylib/builds/1240#annotation-buildifier

dws-uber:
https://github.com/bazelbuild/bazel-skylib/pull/278 should address the CI issues.

dws-uber:
CI is now green!

dws-uber:
Note: I did not make any changes under `docs`. These files appear to be generated. In order to run `docs/regenerate_docs.sh` with bazel 3.1.0 I needed to remove the option `--experimental_remap_main_repo` from that script.  Then, when I ran the script, it did not recreate the docs as they currently exist in that directory. So I'm not sure what the proper process is for updating the files in that directory.

dws-uber:
It occurs to me that the new `types.is_partial()` predicate itself could be considered a reasonable unit of change for a PR.
If you'd prefer to have a separate PR for that, please just let me know.

dws-uber:
@c-parsons @jin @meisterT Any thoughts on this?

lib/types.bzl
```diff
@@ -138,6 +138,20 @@ def _is_set(v):
...
+def _is_partial(v):
+    """Returns True if v is a partial function object created by partial.make().
+
+    Args:
+      v: The value whose type should be checked.
+
+    Returns:
+      True if v was created by partial.make(), False otherwise.
+    """
+    return type(v) == _a_struct_type \
```
alandonovan, line 150: This is only a heuristic. I wonder if making the partial struct an instance of a distinct provider might allow a more robust check.

FWIW, partial is hack to work around the fact that nested def statements don't currently work, but I would like to fix that, at which point we can get start to rid of partial.

lib/types.bzl
```diff
@@ -149,4 +163,5 @@ types = struct(
     is_function = _is_function,
     is_depset = _is_depset,
     is_set = _is_set,
+    is_partial = _is_partial,
```
alandonovan, line 166: Can we move this function to the partial file so that types doesn't depend on it?

lib/types.bzl
dws-uber: I was following the precedent and model of `_is_set()` just above, since it, too, pertains to a skylib-specific type.
Note that `_is_partial()` is a _macro_ and gets used by other macros in skylib, in particular, in `_suite()` in `lib/unittest.bzl`.  My understanding is that providers are available only within _rules_, so they would not be available for use by macros.

Is there an example that you could point me to that would illustrate the alternative implementation you're suggesting?

lib/types.bzl
dws-uber: I'm not sure what you mean by "so that types doesn't depend on it". In my current changes, `is_partial()` in `types.bzl` does not depend on `partial.bzl`, just as `is_set()` does not depend on any of the skylib set implementations.  Both of these predicates do duplicate knowledge of how their respective data structures are implemented elsewhere.

dws-uber (commented):
My goal here is to enable constructing a unittest.suite with rules that have nondefault attributes. Since skylib itself provides for partial instantiation of rules, it seemed reasonable to try to solve the problem using tools that skylib already provides.  But, if there's a different approach that you'd prefer to see, I'll be happy to explore it.

lib/types.bzl
alandonovan: Exactly: the knowledge belongs in sets or partial, not in types. Even though there may be no physical dependency, there is a logical one, and it is unnecessary.

lib/types.bzl
dws-uber: Will do. Would you like the name to remain `is_partial`?  That way the usage sites would look like `partial.is_partial()`, which strikes me as a bit redundant.  Another possibility that occurs to me is to call it `made`, since it checks whether something came from `make`. That way the usage sites would look like `partial.made()`, to check if something came from `partial.make()`.

alandonovan: `partial.is_instance(x)`?  `made` seems a little too cryptic.

dws-uber: Sounds good, I'll use `is_instance()`.

Now, is it okay to make `partial.is_instance()` use the predicates defined in `types.bzl`? Or should I code it so that `partial.bzl` does not do any additional `load()`s?

dws-uber: I will err on the side of being self-contained. If you'd prefer to keep things DRY at the cost of making `partial.bzl` depend on `types.bzl`, please just let me know.

dws-uber: Okay, I believe I've addressed all your comments so far.

lib/partial.bzl
```diff
@@ -124,7 +129,29 @@ def _make(func, *args, **kwargs):
...
+      True if v was created by `make`, False otherwise.
+    """
+    # Note that in bazel 3.7.0 and earlier, type(v.function) is the same
+    # as the type of a function even if v.function is a rule. But we
+    # cannot rely on this in later bazels due to breaking change
+    # https://github.com/bazelbuild/bazel/commit/e379ece1908aafc852f9227175dd3283312b4b82
+    #
+    # Since this check is heuristic anyway, we simply check for the
+    # presence of a "function" attribute without checking its type.
+    return type(v) == _a_struct_type \
```
alandonovan, line 148: It's frustrating that Starlark doesn't offer better features for encapsulation. Like I mentioned, providers act as distinct constructors but don't have an 'instanceof' operation. Another approach would be to add a field whose value is a distinguished value, and check that it matches in is_instance, but there's no way to make the field private to stop external clients from accessing it.

Fortunately in this instance, lambdas should render the whole module unnecessary.

alandonovan (approved):

lib/partial.bzl
dws-uber: I will look forward to better mechanisms in the future. In the meantime, thank you for helping to get this landed!

PR #280, opened 2020-09-29
Fix the docs to match the comment and code.

jkjk822:
Lists are not supported. The code comments were fixed to match this in http://github.com/bazelbuild/bazel-skylib/commit/2d620ba1f8284695181fa092ec4064724ea99a9a, but the docs were never updated.

### DIFF
--- a/docs/new_sets_doc.md
+++ b/docs/new_sets_doc.md
@@ -343,7 +343,7 @@ Returns the union of several sets.
       <td>
         optional.
         <p>
-          An arbitrary number of sets or lists.
+          An arbitrary number of sets.
         </p>
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


jin:
Seems like this should be ran through Stardoc for a complete regeneration?

jkjk822:
Is that something I need to do? Not sure how that is done.

c-parsons:
The docs you are changing in your PR are generated from inline comments in Starlark.

Particularly, here is the line:
https://github.com/bazelbuild/bazel-skylib/blob/2d620ba1f8284695181fa092ec4064724ea99a9a/lib/new_sets.bzl#L185

After the inline comments are changed, please regenerate the docs and include them in the PR. They are the build targets under `docs/`.

jkjk822:
Looks like the docs haven't been regenerated in awhile... The formatting is completely changed.

Maybe the other doc files should be regenerated as well?

c-parsons (approved):

PR #285, opened 2020-12-16
Add paths.globmatch function

erenon:
globmatch(path, pattern) returns True if `path` matches `pattern`.

The "*" wildcard in `pattern` matches zero or more characters.

Examples:

    globmatch("foo", "foo") == True
    globmatch("foobar", "foo") == False
    globmatch("foobar", "foo*") == True
    globmatch("foobar", "f*b*") == True

For performance, the following shortcuts can be added:

```
startcount = pattern.count("*")
if startcount == 0:
  return path == pattern
elif startcount == 1:
  start,_,end = pattern.partition("*")
  return path.startswith(start) and path.endswith(end)
```

### DIFF
--- a/docs/paths_doc.md
+++ b/docs/paths_doc.md
@@ -67,6 +67,53 @@ included, unless omitting them would make the dirname empty.
 </table>
 
 
+## paths.globmatch
+
+<pre>
+paths.globmatch(<a href="#paths.globmatch-path">path</a>, <a href="#paths.globmatch-pattern">pattern</a>)
+</pre>
+
+Returns True if `path` matches `pattern`.
+
+The "*" wildcard in `pattern` matches zero or more characters.
+
+Examples:
+
+    globmatch("foo", "foo") == True
+    globmatch("foobar", "foo") == False
+    globmatch("foobar", "foo*") == True
+    globmatch("foobar", "f*b*") == True
+
+### Parameters
+
+<table class="params-table">
+  <colgroup>
+    <col class="col-param" />
+    <col class="col-description" />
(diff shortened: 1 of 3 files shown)


erenon:
Some CI jobs failed with unrelated "could not download Bazel: HTTP GET https://storage.googleapis.com/bazel-builds/artifacts/macos/8cc0e261a313dbf5e81bfeca2bafa3e12a991046/bazel failed with error 404"

alandonovan:
Thanks Taler, this is very clever. Unfortunately it doesn't belong in skylib. First, I'm not aware there's much call for it. How many projects need this feature besides your own? We should not publish things in skylib until there is a demonstrated widespread compelling need, and the interface is unimpeachable. (Few of the declarations already in skylib would pass this bar, but that is not a reason to add to them.)

Second, globmatch doesn't match glob. The semantics of Bazel's built-in glob function are rather more complex, and it would be quite tricky to simulate them accurately in Starlark, and even harder to do so efficiently. Anything less is a liability.

If this function is useful to you, I suggest you put it your own project. If another project needs it, they can copy it. If many projects need it, you could publish it in a utility repo.

erenon:
Thanks for the quick reply.

> We should not publish things in skylib until there is a demonstrated widespread compelling need

I did not find this in the README.md or in the CONTRIBUTING.md, while there's a short doc in the readme on how to add new modules. Perhaps it would be good to mention it somewhere, to avoid others wasting time.

alandonovan:
Good idea. https://github.com/bazelbuild/bazel-skylib/pull/286/files

PR #286, opened 2020-12-16
Explain criteria for contributing

alandonovan:


README.md
```diff
@@ -88,7 +102,7 @@ Steps to add a module to Skylib:
    ```
 
 1. Add unit tests for your module in the `tests` directory.
-
+w
```
thomasvl, line 105: stray?

README.md
```diff
@@ -66,6 +65,21 @@ s = shell.quote(p)
 
 ## Writing a new module
 
+The criteria for adding a new function or module to this repository are:
+
+1. Is it widely needed? The new code must solve a problem that occurs often during the development of Bazel build rules. It is not sufficient that the new code is merely useful.  Candidate code should generally have been proven to be indispensible in the repository in which it was developed, or across several projects.
```
juliexxia, line 70: "Indispensible" strikes me as too strong of wording. For example, the build settings in rules/common_settings.bzl were created at the inception of build settings because we knew it would prevent the implementation of many extremely-similar build settings across projects and allowed us to get simple ones out there that used best practices.

So IMO another reasonable reason to add content is because it's going to steer users towards a more standardized way to use features which is known to work and will save them implementation time. WDYT?

README.md
```diff
@@ -2,12 +2,11 @@
...
 
-Skylib is a standard library that provides functions useful for manipulating
-collections, file paths, and other features that are useful when writing custom
-build rules in Bazel.
+Skylib is a library of Starlark functions for manipulating collections, file paths,
+and various other data types in the domain of Bazel build rules.
 
-> This library is currently under early development. Be aware that the APIs
-> in these modules may change during this time.
+This library is currently under early development. Be aware that the APIs
```
aiuto, line 8: Perhaps it is time to change this statement.

It seems to me like the motivation behind this PR is that things added here essentially live forever, so we have to be careful about what is added. That is at odds with fast changing APIs.

alandonovan: Indeed. Removed.

README.md
alandonovan: Rephrased.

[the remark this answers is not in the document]
alandonovan: Good point. I've made that the emphasis.

README.md
alandonovan: Oops, removed.

aiuto (approved):

PR #287, opened 2020-12-16
Add a helper getter method for common settings

juliexxia:


gregestren (approved):

gregestren:
I think you need OWNERS approval from someone else?

juliexxia:
Dang, I thought I was an owner to be honest.

rules/common_settings.bzl
```diff
@@ -98,3 +98,7 @@ string_setting = rule(
     },
     doc = "A string-typed build setting that cannot be set on the command line",
 )
+
+# Extracts the value from an attribute that provides BuildSettingInfo
```
alandonovan, line 102: My review of your PR is another PR:
https://github.com/bazelbuild/bazel-skylib/pull/286/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R68
:)

alandonovan (changes requested):

rules/common_settings.bzl
juliexxia: Not sure what the implication is here - are you saying you don't think this new content meets the criteria you laid out?

aiuto:
ping. lets resolve this.

rules/common_settings.bzl
alandonovan: My point is that this is not a good abstraction, and we certainly wouldn't want users to declare a dependency on this repo for a function like this. It neither hides concepts from the user, nor simplifies the cognitive load. ("The doc comment is longer than the function body." 

(It doesn't even save the user keystrokes: the function's name is longer than it body!)

gregestren: I think @juliexxia had a newer perspective on this PR. Let's hold off more debate until she can offer her input.

PR #288, opened 2021-01-28
Add shebang lines to tests directory shell scripts

olekw:
Prevent Debian's lintian from complaining about scripts without shebangs. There does not seem to be a reason (that I can see) for the shebang being omitted. Since it is present in other shell scripts in the repository, I'm assuming that it's just an oversight that it's missing on these three files.


tetromino (approved):

tetromino:
Thanks.

PR #297, opened 2021-05-11
Run regenerate_docs.sh

alexeagle:
Seems that recent edits were done by hand? Is the generator script still functioning?

- remove flag that no longer exists in recent bazel releases from regenerate_docs script
- use a less verbose path in the output tree so bazel-bin doesn't look so strange
- now uses markdown tables since that's what the pinned version of stardoc does


alexeagle:
Looks like the problem was reported last time someone touched the docs, but then not followed up? 
https://github.com/bazelbuild/bazel-skylib/pull/280#issuecomment-705843592

docs/regenerate_docs.sh
```diff
@@ -19,9 +19,9 @@
 
 set -euo pipefail
 
-bazel build docs:all --experimental_remap_main_repo
+bazel build docs:all
 
-for filename in bazel-bin/docs/*_gen.md; do
```
tetromino, line 24: I would prefer to avoid changing the rename logic here (and thus avoid changing all the `out`s in docs/BUILD) without a specific reason. The existing logic works, and "_gen" suffix in a generated doc file is a good signal.

docs/BUILD
```diff
@@ -6,132 +6,132 @@ licenses(["notice"])
 # the --incompatible_remap_main_repo flag.
 stardoc(
     name = "build_test_docs",
-    out = "build_test_doc_gen.md",
```
tetromino, line 9: I don't see a good reason for the `out` renames here and below in docs/BUILD. "_gen" means a generated, not hand-written md doc.

tetromino (commented):
Thank you for catching this! Agree with the PR, but I would like to limit it just to regenerating the docs and removing the obsolete --experimental_remap_main_repo flag.

docs/BUILD
alexeagle: sure, I'll open a separate PR for that change and discuss my reasoning.

alexeagle:
@tetromino is the intent for the docs to always be up-to-date? If so, the root cause here was a missing `diff_test` which I could add (similar to recently added to rules_apple for instance https://github.com/bazelbuild/rules_apple/blob/master/doc/BUILD.bazel#L36-L45 )

alexeagle:
ping @tetromino would you like tests added here that assert the docs are always up-to-date?

tetromino:
@alexeagle - yes, a test to check docs are up to date would be greatly appreciated (as a follow-up PR).

tetromino (approved):
Thanks, looks good!

There are some conflicts reported - let me see if they can be fixed trivially...

Issue #189, opened 2019-09-04
Test rule to verify aspect applied to another target


c-parsons:
Some workflows apply a top level aspect to a build, such as:

```
bazel build //foo:bar --aspect=//baz:my_aspect.bzl%my_aspect
```

It would be nice to have a canonical test rule which would verify that builds such as these would work.

It should be as simple as creating a thin analysis test rule which applies the aspect along the edge. That is, a test rule with attribute defined:

```
deps' : attr.label_list(aspects = [my_aspect]),
```
And then, to satisfy the above example, the user could create a test target that has `//foo:bar` as one of its `deps`.

PR #299, opened 2021-05-24
Allow specifying additional aspects to tut

illicitonion:
To simulate this in a test I had to define a separate attribute with the
aspect applied and ignore `target_under_test`, except because
`target_under_test` is mandatory I had to set some value for it which
was ignored.

This may or may not fix
https://github.com/bazelbuild/bazel-skylib/issues/189 (I could see also
wanting a `build_test`-like target which builds with an aspect - I'm not
entirely sure what the scope of that issue was meant to be :))

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -151,7 +151,8 @@ def _make_analysis_test(
         expect_failure = False,
         attrs = {},
         fragments = [],
-        config_settings = {}):
+        config_settings = {},
+        extra_target_under_test_aspects = []):
     """Creates an analysis test rule from its implementation function.
 
     An analysis test verifies the behavior of a "real" rule target by examining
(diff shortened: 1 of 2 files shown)


hlopko:
+1, without this PR or other solution to #189 I am not able to conveniently unit test aspects. My workaround is to create an artificial rule that applies the aspect on its attribute, and collects the aspect outputs, but I'm still not able to make assertions about providers provided by the aspect on targets it's attached to.

I'm happy to write up a longer problem description if ^ is not clear enough.

illicitonion:
@tetromino - would you mind taking a quick look? Thanks!

UebelAndre:
@brandjon @tetromino would you be able to take a look here? 😅 🙏

tetromino (changes requested):
Looks good, but can you please add a small test to tests/unittest_tests.bzl and unittest_test.sh?

tetromino (approved):
Thanks for the test!

PR #302, opened 2021-06-02
Windows support for build_test

stdbug:
I have to point out that the approach used in this PR is a bit hacky as it marks `.bat` file as executable on Windows. AFAIK, in order to spawn processes on Windows, Bazel uses `CreateProcessW` function whose [docs](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) says that caller should use `cmd.exe` as the application name and the batch file as one of `lpCommandLine` arguments. On the other hand, I don't know any Windows flavor that wouldn't run the batch file if it's passed as the application name to `CreateProcessW`

### DIFF
--- a/rules/build_test.bzl
+++ b/rules/build_test.bzl
@@ -17,16 +17,13 @@
 load("//lib:new_sets.bzl", "sets")
 
 def _empty_test_impl(ctx):
-    is_windows = select({
-        "@bazel_tools//src/conditions:host_windows": True,
-        "//conditions:default": False,
-    })
-    extension = ".bat" if is_windows else ".sh"
+    extension = ".bat" if ctx.attr.is_windows else ".sh"
+    content = "exit 0" if ctx.attr.is_windows else "#!/bin/bash\nexit 0"
     executable = ctx.actions.declare_file(ctx.label.name + extension)
     ctx.actions.write(
         output = executable,
         is_executable = True,
-        content = "exit 0",
+        content = content,
     )
 
     return [DefaultInfo(
@@ -39,6 +36,7 @@ _empty_test = rule(
     implementation = _empty_test_impl,
     attrs = {
         "data": attr.label_list(allow_files = True),
+        "is_windows": attr.bool(mandatory = True),
     },
     test = True,
 )
@@ -108,5 +106,9 @@ def build_test(name, targets, **kwargs):
         name = name,
         data = test_data,
         size = kwargs.pop("size", "small"),  # Default to small for test size
+        is_windows = select({
+            "@bazel_tools//src/conditions:host_windows": True,
+            "//conditions:default": False,
+        }),
         **kwargs
     )
--- a/rules/build_test.bzl
+++ b/rules/build_test.bzl
@@ -16,15 +16,43 @@
 
 load("//lib:new_sets.bzl", "sets")
 
+def _empty_test_impl(ctx):
+    is_windows = select({
+        "@bazel_tools//src/conditions:host_windows": True,
+        "//conditions:default": False,
+    })
+    extension = ".bat" if is_windows else ".sh"
+    executable = ctx.actions.declare_file(ctx.label.name + extension)
+    ctx.actions.write(
+        output = executable,
+        is_executable = True,
+        content = "exit 0",
+    )
+
+    runfiles = []
+    for entry in ctx.attr.data:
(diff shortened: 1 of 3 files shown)


rules/build_test.bzl
```diff
@@ -16,16 +16,44 @@
...
+    return [DefaultInfo(
+        files = depset([executable]),
+        executable = executable,
+        runfiles = ctx.runfiles(files = runfiles),
+    )]
+
+_empty_test = rule(
+    implementation = _empty_test_impl,
+    attrs = {
+        "data": attr.label_list(),
```
FaBrand, line 45:
```suggestion
        "data": attr.label_list(allow_files = True),
```
You can use the `allow_files` to make the code above simpler. No need to [call to_list() in a loop](https://docs.bazel.build/versions/master/skylark/depsets.html#performance)

rules/build_test.bzl
```diff
@@ -16,16 +16,44 @@
...
+    )
+
+    runfiles = []
+    for entry in ctx.attr.data:
+        runfiles += entry.files.to_list()
+
+    return [DefaultInfo(
+        files = depset([executable]),
+        executable = executable,
+        runfiles = ctx.runfiles(files = runfiles),
```
FaBrand, line 39:
```suggestion
        runfiles = ctx.runfiles(files = ctx.files.data),
```

rules/build_test.bzl
```diff
@@ -16,16 +16,44 @@
...
+    ctx.actions.write(
+        output = executable,
+        is_executable = True,
+        content = "exit 0",
+    )
+
+    runfiles = []
+    for entry in ctx.attr.data:
+        runfiles += entry.files.to_list()
+
```
FaBrand, line 35:
```suggestion
```
Not necessary with `ctx.files.data`

FaBrand (commented):
Some suggestions to make the empty test more concise with its runfiles

rules/build_test.bzl
```diff
@@ -16,16 +16,40 @@
 
 load("//lib:new_sets.bzl", "sets")
 
+def _empty_test_impl(ctx):
+    is_windows = select({
+        "@bazel_tools//src/conditions:host_windows": True,
+        "//conditions:default": False,
+    })
+    extension = ".bat" if is_windows else ".sh"
```
FaBrand, line 24:
```suggestion
    extension = ".bat" if ctx.attr.is_windows else ".sh"
```

rules/build_test.bzl
```diff
@@ -16,16 +16,40 @@
...
+    return [DefaultInfo(
+        files = depset([executable]),
+        executable = executable,
+        runfiles = ctx.runfiles(files = ctx.files.data),
+    )]
+
+_empty_test = rule(
+    implementation = _empty_test_impl,
+    attrs = {
+        "data": attr.label_list(allow_files = True),
```
FaBrand, line 41:
```suggestion
        "data": attr.label_list(allow_files = True),
        "is_windows": attr.bool(mandatory = True),
```

rules/build_test.bzl
```diff
@@ -75,15 +99,13 @@ def build_test(name, targets, **kwargs):
...
             **genrule_args
         )
 
-    native.sh_test(
+    _empty_test(
         name = name,
-        # TODO: Does this need something else for Windows?
-        srcs = ["@bazel_skylib//rules:empty_test.sh"],
         data = test_data,
         size = kwargs.pop("size", "small"),  # Default to small for test size
```
FaBrand, line 110:
```suggestion
        size = kwargs.pop("size", "small"),  # Default to small for test size
        is_windows = select({
                "@bazel_tools//src/conditions:host_windows": True,
                "//conditions:default": False,
        }),
```

rules/build_test.bzl
FaBrand: This is a bug, the select is not evaluated as desired in this context.
However it is truthy, therfore the if does parse. Therefore the condition always evaluates to `True` and extension .bat is always used.

rules/build_test.bzl
FaBrand: This is evaluatited in the `BUILD`, file, therefore the select is resolved as expected.

meteorcloudy (approved):

tetromino (approved):

PR #307, opened 2021-06-20
diff_test: add ability for caller to specify a message printed on fai…

alexeagle:
…lure.

Useful for the use case of a pair of .test and .update targets for checked-in golden files


keith (approved):
awesome, we've wanted this to so we can advise on things like "run this command to fix this"

keith:
@brandjon @tetromino can you take a look?

tetromino (approved):
Thanks!

PR #308, opened 2021-07-14
Update partial_doc.md

dmaclach:
Clean up the docs to make them look significantly better.

No semantics changes, just styling.

PR #309, opened 2021-07-15
Update partial.bzl to generate better markdown

dmaclach:
For https://github.com/bazelbuild/bazel-skylib/pull/308 I edited the markdown directly. I should've updated the bzl file and generated the documentation.


tetromino (changes requested):
Buildkite is complaining about buildifier errors; please run `buildifier lib/partial.bzl`

dmaclach:
Updated

tetromino (approved):

PR #311, opened 2021-08-09
Remove `$` from grep regexes

philwo:
Due to grep having dropped support for handling line-ending matches in a cross-platform way, grepping for `...$` will now fail on Windows, as it no longer ignores the CR part of the CRLF line endings on Windows.

This should turn this project green again on Bazel CI.


meteorcloudy (approved):
Thanks!

tetromino (approved):
Thanks!

PR #314, opened 2021-08-16
Make diff_test work on windows with --enable_runfiles

### LINKED ISSUE #313: diff_test fails on windows with --enable_runfiles
To reproduce, on windows:

    > bazel test tests/diff_test/...  --enable_runfiles

Root cause:

```
set MF=%RUNFILES_MANIFEST_FILE:/=\\%
set PATH=%SYSTEMROOT%\\system32
set F1={file1}
set F2={file2}
if "!F1:~0,9!" equ "external/" (set F1=!F1:~9!) else (set F1=!TEST_WORKSPACE!/!F1!)
if "!F2:~0,9!" equ "external/" (set F2=!F2:~9!) else (set F2=!TEST_WORKSPACE!/!F2!)
for /F "tokens=2* usebackq" %%i in (`findstr.exe /l /c:"!F1! " "%MF%"`) do (
  set RF1=%%i
  set RF1=!RF1:/=\\!
)
```

https://github.com/bazelbuild/bazel-skylib/blob/main/rules/diff_test.bzl#L36

The batch script expects that a MANIFEST file exists, but it does not, if `--enable_runfiles` is set, even if `--build_runfiles_manifests` is also set.

erenon:
If `--enable_runfiles` is set, no manifest file will be created, and `RUNFILES_MANIFEST_FILE` will be empty,
but the runfiles will be accessible from the current directory using their canonical names.
If a file cannot be found in the manifest file, check if it can be found relative from pwd.
This is the same logic as in the shell counterpart around line 100.

Fixes #313

### DIFF
--- a/rules/diff_test.bzl
+++ b/rules/diff_test.bzl
@@ -44,16 +44,26 @@ for /F "tokens=2* usebackq" %%i in (`findstr.exe /l /c:"!F1! " "%MF%"`) do (
   set RF1=!RF1:/=\\!
 )
 if "!RF1!" equ "" (
-  echo>&2 ERROR: !F1! not found
-  exit /b 1
+  if exist "{file1}" (
+    set RF1="{file1}"
+    set RF1=!RF1:/=\\!
+  ) else (
+    echo>&2 ERROR: !F1! not found
+    exit /b 1
+  )
 )
 for /F "tokens=2* usebackq" %%i in (`findstr.exe /l /c:"!F2! " "%MF%"`) do (
   set RF2=%%i
   set RF2=!RF2:/=\\!
 )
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):

meteorcloudy:
This change looks good. 

An alternative is to locate the MANIFEST file with similar logic here:
https://github.com/bazelbuild/bazel/blob/29e44fecf71fe45b1d9fdeb3ff1b8e379a3df9e0/tools/bash/runfiles/runfiles.bash#L71. It should be generated even if --enable_runfiles is turned on.

tetromino (approved):

PR #321, opened 2021-10-09
Add diff_test asserting that docs are up-to-date

alexeagle:
it prints a convenient 'bazel run' command to update them, replacing the shell script

Follow-up to https://github.com/bazelbuild/bazel-skylib/pull/297#issuecomment-926693513


alexeagle:
I agree in theory that BUILD files should be machine-edited.

However in this case, there is no likely machine-editor in sight. This repo isn't big enough to require one-off `buildozer` commands, and there is no gazelle plugin to help maintain the file.

If it's always going to be hand-edited, would you still want to expand the file to the amount of boilerplate you suggest?

In addition the overall complexity is higher when you need to create that macro and (very cool) way of finding other rules in the package.

OTOD you've already got the edits in context so WDYT about making that change yourself?

alexeagle:
Hey @tetromino in the time since you suggested that, we built a full-featured rule to support exactly your proposal
https://github.com/aspect-build/bazel-lib/blob/main/docs/docs.md
example usage:
https://github.com/aspect-build/rules_js/blob/main/docs/BUILD.bazel

tetromino:
@alexeagle - the aspect_build/bazel-lib rule looks great for skylib's internal use.

I'm hesitant to export it as is a public api from skylib because compared to a Google internal implementation of the same idea, it's missing some features which non-Google users probably also want (e.g. autoformatting of markdown output, keeping docs in a separate directory tree from .bzl sources, embedding comments in the markdown output), and the "package:" tag feels like bit of a hack.

Are you interested in updating this PR based on aspect_build/bazel-lib stardoc_with_diff_test without the "package:" tag (which isn't needed for internal functionality)?

tetromino:
Or alternatively, do you want me to do it?

alexeagle:
Sure, I can do it. I'd really like to have more collaboration with you, I think we are doing a lot of the same work.

https://github.com/aspect-build/bazel-lib/blob/6f7469dd03dd1a19697a3d7e33d7ba8972eac034/lib/private/docs.bzl has the implementation you want, prior to switching it to be based on our `write_source_file` rule. That rule has the nice property that it gives the user a nice error when the source file to diff against doesn't exist, but it's not a feature you need.

/cc @kormide

alexeagle:
ping @tetromino PTAL

tetromino (approved):
LGTM and thank you!

I took the liberty to fix a couple of tiny nits - added some comments to remind that update_docs needs to go at the bottom of the BUILD file, and renamed the tag to follow the same naming scheme as other projects.

PR #323, opened 2021-10-09
copy_file allows directories as well

### LINKED ISSUE #322: copy_file doesn't work for directories
There should be a way to copy directories, just like we copy files.

It would use ctx.actions.declare_directory so that Bazel tracks the result as a TreeArtifact.

We need this in rules_nodejs for example, because third-party packages are always on disk as directories, and globbing them as a bunch of files runs into filename restrictions (spaces in names) and is very non-performant (thousands of files to symlink into execroot/runfiles trees).

Another reasoning for this: RBE doesn't allow InputArtifacts that are directories (see @ulfjack's note on https://github.com/bazelbuild/bazel/commit/c64421bc35214f0414e4f4226cc953e8c55fa0d2) so it's necessary to copy them as TreeArtifacts to use them in remote-compatible actions.

alexeagle:
Fixes #322

### DIFF
--- a/rules/private/copy_file_private.bzl
+++ b/rules/private/copy_file_private.bzl
@@ -14,11 +14,21 @@
 
 """Implementation of copy_file macro and underlying rules.
 
-These rules copy a file to another location using Bash (on Linux/macOS) or
-cmd.exe (on Windows). '_copy_xfile' marks the resulting file executable,
-'_copy_file' does not.
+These rules copy a file or directory to another location using Bash (on Linux/macOS) or
+cmd.exe (on Windows). `_copy_xfile` marks the resulting file executable,
+`_copy_file` does not.
 """
 
+# Hints for Bazel spawn strategy
+_execution_requirements = {
+    # Copying files is entirely IO-bound and there is no point doing this work remotely.
+    # Also, remote-execution does not allow source directory inputs, see
+    # https://github.com/bazelbuild/bazel/commit/c64421bc35214f0414e4f4226cc953e8c55fa0d2
+    # So we must not attempt to execute remotely in that case.
+    "no-remote-exec": "1",
+}
+
+# buildifier: disable=function-docstring
 def copy_cmd(ctx, src, dst):
     # Most Windows binaries built with MSVC use a certain argument quoting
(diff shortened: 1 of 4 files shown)


lencioni:
@brandjon @tetromino do you have any feedback on this PR? I believe this will help unblock using rules_nodejs with remote build execution, which would be really nice to see!

alexeagle:
Ping! Is anyone triaging PRs in this repo?
If not we'll probably switch rules_nodejs to use https://github.com/aspect-build/bazel-lib where we can get fixes like this.

rules/private/copy_file_private.bzl
```diff
@@ -14,23 +14,46 @@
...
     # cmd.exe uses different semantics, so Bazel's quoting is wrong here.
     # To fix that we write the command to a .bat file so no command line
     # quoting or escaping is required.
     bat = ctx.actions.declare_file(ctx.label.name + "-cmd.bat")
+
+    # Flags are documented at
+    # https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/copy
+    # https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/xcopy
+    if dst.is_directory:
+        cmd_tmpl = "@xcopy \"%s\" \"%s\\\" /V /E /H /Y /Q >NUL"
```
tetromino, line 44: Why use /v for xcopy, but not for copy?

Why delete the destination directory in the bash version, but not in the Windows version?

tetromino (changes requested):

tetromino (changes requested):
Overall - looks good; thank you for the patch!

However, I would prefer a separate copy_directory rule: less confusing name (files aren't directories), no confusing interaction between is_directory and allow_symlinks flags, more noticeable documentation of DBAZEL_TRACK_SOURCE_DIRECTORIES.

The one new test is insufficient. Scenarios we'd probably want to test:
* copying an empty directory
* copying a directory whose content includes a symlink
* copying a directory to a location which is an existing file
* copying a directory to a location which is an existing directory

alexeagle:
replaced by #366

PR #324, opened 2021-10-09
fix: allow multiple copy_file in the same package

alexeagle:
Previously this failed on windows, as we wrote colliding batch files

### DIFF
--- a/rules/private/copy_file_private.bzl
+++ b/rules/private/copy_file_private.bzl
@@ -19,13 +19,19 @@ cmd.exe (on Windows). '_copy_xfile' marks the resulting file executable,
 '_copy_file' does not.
 """
 
+def _hash_file(file):
+    return str(hash(file.path))
+
 def copy_cmd(ctx, src, dst):
     # Most Windows binaries built with MSVC use a certain argument quoting
     # scheme. Bazel uses that scheme too to quote arguments. However,
     # cmd.exe uses different semantics, so Bazel's quoting is wrong here.
     # To fix that we write the command to a .bat file so no command line
     # quoting or escaping is required.
-    bat = ctx.actions.declare_file(ctx.label.name + "-cmd.bat")
+    # Put a hash of the file name into the name of the generated batch file to
+    # make it unique within the package, so that users can define multiple copy_file's.
+    bat = ctx.actions.declare_file("%s-%s-cmd.bat" % (ctx.label.name, _hash_file(src)))
+
     ctx.actions.write(
         output = bat,
         # Do not use lib/shell.bzl's shell.quote() method, because that uses


alexeagle:
Ping @tetromino

rules/private/copy_file_private.bzl
```diff
@@ -19,13 +19,19 @@ cmd.exe (on Windows). '_copy_xfile' marks the resulting file executable,
...
+    return str(hash(file.path))
+
 def copy_cmd(ctx, src, dst):
     # Most Windows binaries built with MSVC use a certain argument quoting
     # scheme. Bazel uses that scheme too to quote arguments. However,
     # cmd.exe uses different semantics, so Bazel's quoting is wrong here.
     # To fix that we write the command to a .bat file so no command line
     # quoting or escaping is required.
-    bat = ctx.actions.declare_file(ctx.label.name + "-cmd.bat")
+    # Put a hash of the file name into the name of the generated batch file to
```
alexeagle, line 31: note: users of the public API wouldn't observe a problem because only one declare_file happens per `copy_file` rule, and so `ctx.label.name` is sufficient to disambiguate.

We hit this in a scenario where a rule uses this library code and creates multiple actions.

tetromino (commented):
I might be blind here, but I don't think I understand the purpose of this change.

As far as I can see, two `copy_file` targets (even with the same src) in the same package shouldn't collide on Windows because they will have different labels.

It seems the fix is only for a hypothetical situation where another rule's implementation in Skylib might call copy_cmd in a loop - but no such rule exists at the moment (again, as far as I can see).

Is this a prerequisite for another PR - if so, which one? I don't think that #323 should need this...

alexeagle:
Yes I mentioned it in https://github.com/bazelbuild/bazel-skylib/pull/324#discussion_r840723348
the particular rule in this case is https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/js_library/js_library.bzl#L34-L38

Granted, that rule uses the private API of `copy_file` and therefore there's no support guarantee for doing this. You could just say no. However I think it's nice to fix it anyway.

tetromino:
I'm wondering if we should instead bite the bullet and fix `run_shell` to support .bat and powershell commands. Then we wouldn't need to write temporary bat files at all, here and elsewhere.

Wrote up my thoughts in https://github.com/bazelbuild/bazel/issues/15194

alexeagle:
That looks interesting, but it's totally orthogonal to this fix, right?

tetromino:
If `run_shell` supported `command_bat`, we wouldn't need copy_cmd and copy_bat at all: the implementation would be unified into

```
    args = ctx.actions.args()
    args.add_all(src.path, dst.path)
    ctx.actions.run_shell(
        tools = [src],
        outputs = [dst],
        command = "cp -f \"$1\" \"$2\"",
        command_bat = "copy /Y \"%1\" \"%2\" >NUL",
        arguments = [args],
        mnemonic = "CopyFile",
        progress_message = "Copying files",
        use_default_shell_env = True,
    )
```

PR #326, opened 2021-10-16
copy_file: Do not add non-executables to default_runfiles

fmeum:
copy_file currently includes the copied file in its runfiles even if it is not executable, which makes every rule depending on it have the file as a runfile (e.g. a `cc_library` depending on a copied header file via the hdrs attribute).

In an ideal world, according to https://docs.bazel.build/versions/main/skylark/rules.html#runfiles-features-to-avoid, `copy_file` would not need to specify any runfiles in the `DefaultInfo` it returns - specifying `files` should suffice. However, due to the existence of rules with legacy behavior, this would break compatibility (actually, already with `sh_test` in skylib's unit tests).

As a compromise that preserves compatibility with legacy rules but nevertheless cleans up the runfiles tree of depending rules, use the `data_runfiles` attribute of `DefaultInfo` if the copied file is not executable.

(Note: I noticed the problem with legacy rules only after submitting https://github.com/bazelbuild/bazel-skylib/pull/325, so please ignore that one).

### DIFF
--- a/rules/private/copy_file_private.bzl
+++ b/rules/private/copy_file_private.bzl
@@ -75,7 +75,10 @@ def _copy_file_impl(ctx):
     if ctx.attr.is_executable:
         return [DefaultInfo(files = files, runfiles = runfiles, executable = ctx.outputs.out)]
     else:
-        return [DefaultInfo(files = files, runfiles = runfiles)]
+        # Do not include the copied file into the default runfiles of the
+        # target, but ensure that it is picked up by native rule's data
+        # attribute despite https://github.com/bazelbuild/bazel/issues/15043.
+        return [DefaultInfo(files = files, data_runfiles = runfiles)]
 
 _ATTRS = {
     "src": attr.label(mandatory = True, allow_single_file = True),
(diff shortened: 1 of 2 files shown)


tetromino (changes requested):
Seems reasonable, but shouldn't we make the same change to write_file too?

fmeum:
@tetromino I have learned more about runfiles in the meantime and now believe that this change is not net positive: The only reason that setting runfiles is necessary at all in this case is the legacy behavior of native rules caused by https://github.com/bazelbuild/bazel/issues/15043. I now think that it would make more sense to wait for the fix to land and at some point drop the `runfiles` attribute entirely from these rules instead of using the deprecated `data_runfiles` as a band aid fix in the meantime. Please let me know if you think differently, I'm still not sure I have a full grasp on the idiomatic way of specifying runfiles.

tetromino:
I'm neutral on this - but note that stable Bazel users won't see the fix to https://github.com/bazelbuild/bazel/pull/15052 until Bazel 7.

fmeum:
@tetromino Good point. I made the change also for `write_file` and added (hopefully) helpful comments in both places.

tetromino (approved):
Thanks!

PR #338, opened 2022-01-05
Fix name arg documentation for native_binary

GMNGeoffrey:


GMNGeoffrey:
Side note: why does this need to copy the binary instead of a symlinking? The version I have of this rule appears to work fine as a symlink

GMNGeoffrey:
@c-parsons could you PTAL

tetromino (approved):
Thanks!

PR #339, opened 2022-01-05
Fix runfiles in native_binary/native_test

GMNGeoffrey:
The current implementation misses the runfiles from the binary itself
since the attribute for it is called `src` not `srcs` and it makes use
of the discouraged `collect_data` and `collect_default` parameters which
depend on hardcoded attribute names.


GMNGeoffrey:
@c-parsons could you PTAL

rules/native_binary.bzl
```diff
@@ -28,14 +28,17 @@ def _impl_rule(ctx, is_windows):
         copy_cmd(ctx, ctx.file.src, out)
     else:
         copy_bash(ctx, ctx.file.src, out)
+    runfiles = ctx.runfiles(files = ctx.files.data)
+
+    # Not using merge_all because it is not available in Bazel LTS 4.x
```
tetromino, line 33: We should use `merge_all` if it is available; Bazel 5 LTS is coming soon. Something like this:

```
    # Bazel 4.x LTS does not support `merge_all`.
    # TODO: remove `merge` branch once we drop support for Bazel 4.x.
    if hasattr(runfiles, "merge_all"):
        runfiles = runfiles.merge_all([
            d[DefaultInfo].default_runfiles
            for d in ctx.attr.data + [ctx.attr.src]
        ])
    else:
        ...
```

tetromino (changes requested):
Please add a test to verify that runfiles from the binary's src are collected.

rules/native_binary.bzl
GMNGeoffrey: Done

GMNGeoffrey:
> Please add a test to verify that runfiles from the binary's src are collected.

Done. Also fixed issues with the existing tests. This test fails at tot and passes with my PR.

GMNGeoffrey:
Ping on this?

GMNGeoffrey:
@tetromino ping on this again

GMNGeoffrey:
Hello there! This has been open for nearing on 3 months here. Could someone please take a look? I've got two other PRs on this same rule (#340 and #341) and have made even further enhancements to it in the fork we're using for our own project. The pace of review here is extremely disappointing

tetromino (approved):
Sorry for the disappointing pace of review. Thanks for fixing the tests!

PR #340, opened 2022-01-05
Switch native_binary/test to use symlinks

GMNGeoffrey:
A copy should not be necessary here. The symlink action falls back to
copying when symlinks are not enabled on windows, which I think would 
be the only problem with using symlinks:
https://docs.bazel.build/versions/4.2.2/command-line-reference.html#flag--windows_enable_symlinks


GMNGeoffrey:
Depends on https://github.com/bazelbuild/bazel-skylib/pull/339, so only the last commit is part of this PR (anyone's who's figured out a good way to have chains of PRs with GitHub please let me know).

GMNGeoffrey:
@c-parsons could you PTAL

rules/native_binary.bzl
```diff
@@ -20,14 +20,13 @@ do, but they run the wrapped binary directly, instead of through Bash, so they
 don't depend on Bash and work with --shell_exectuable="".
```
tetromino, line 20: Nit: take the opportunity to fix misspelling? `--shell_executable`, not exectuable

rules/native_binary.bzl
```diff
@@ -48,79 +47,39 @@ def _impl_rule(ctx, is_windows):
...
-
 _ATTRS = {
     "src": attr.label(
         executable = True,
         allow_single_file = True,
         mandatory = True,
         cfg = "host",
+        doc = "path of the pre-built executable",
     ),
     "data": attr.label_list(allow_files = True),
```
tetromino, line 58: Nit: `data` attr needs a docstring since the macros and the macros' docs are gone

rules/native_binary.bzl
```diff
@@ -20,14 +20,13 @@ do, but they run the wrapped binary directly, instead of through Bash, so they
 don't depend on Bash and work with --shell_exectuable="".
 """
 
-load("//rules/private:copy_file_private.bzl", "copy_bash", "copy_cmd")
```
tetromino, line 23: Since we're no longer loading copy_bash/copy_cmd, please remove the dep on //rules/private:copy_file_private in the //rules:native_binary bzl_library

tetromino (changes requested):
Overall - nice simplification!

Note that tests are failing because docs need to be regenerated (bazel build //docs:update)

rules/native_binary.bzl
GMNGeoffrey: Done

rules/native_binary.bzl
GMNGeoffrey: Done

rules/native_binary.bzl
GMNGeoffrey: Done

GMNGeoffrey (commented):
> bazel build //docs:update

Done. Note that it's actually `bazel run //docs:update`

rules/native_binary.bzl
```diff
@@ -17,17 +17,16 @@
...
-
-def _impl_rule(ctx, is_windows):
+def _impl_rule(ctx):
     out = ctx.actions.declare_file(ctx.attr.out)
-    if is_windows:
-        copy_cmd(ctx, ctx.executable.src, out)
-    else:
-        copy_bash(ctx, ctx.executable.src, out)
+    ctx.actions.symlink(
+        target_file = ctx.file.src,
```
tetromino, line 26: From https://buildkite.com/bazel/bazel-skylib/builds/2115#a7024d75-d3b5-4cff-a048-c207efbb6c0b:

```
ERROR: /workdir/tests/native_binary/BUILD:100:12: in native_test rule //tests/native_binary:data_test:
Traceback (most recent call last):
	File "/Users/arostovtsev/bazelbuild/bazel-skylib/rules/native_binary.bzl", line 26, column 31, in _impl_rule
		target_file = ctx.file.src,
Error: No attribute 'src' in file. Make sure there is a label type attribute marked as 'allow_single_file' with this name
```

Note that https://github.com/bazelbuild/bazel-skylib/pull/341 changed `src` to be `allow_files`.

tetromino (changes requested):

rules/native_binary.bzl
GMNGeoffrey: Whoops. Missed that when merging. Fixed

tetromino (approved):
LGTM, thanks!

PR #341, opened 2022-01-06
Build native_binary/test src in correct configuration

GMNGeoffrey:
This attribute is incorrectly being built in the host configuration when
(like any test) it will run in the target configuration. This means that
cross compilation will be broken and options that differ between host
and target (e.g. `NDEBUG`) will not be as set by the user.

I confirmed that without this fix, a test binary with `assert(false)`
passes when run under `native_test`.

Additionally, the use of `allow_single_file` precludes rules that return
multiple files in their DefaultInfo (like `py_binary`). Instead, we can
use `allow_files` and access via `ctx.executable`.


GMNGeoffrey:
@c-parsons could you PTAL

GMNGeoffrey:
> @c-parsons could you PTAL

Ah apologies I think this ping (and similar on my other PRs) was perhaps misdirected. You're identified as the best reviewer internally, but I think that's just because you imported changes from others. I would still appreciate review (or at least acknowledgement) from someone on these. These issues are blocking other things I'm working on

tetromino (approved):
Thanks for the fix!

PR #343, opened 2022-01-07
Allow unit test rules to be documented

UebelAndre:
I find it more ergonomic to add notes or a description about a test to a `doc` attribute, similar to other rules, vs a comment block above it or in the `impl` function docstring. Hopefully this can improve the readability of test rules.

### DIFF
--- a/docs/unittest_doc.md
+++ b/docs/unittest_doc.md
@@ -37,7 +37,7 @@ unittest_toolchain(<a href="#unittest_toolchain-name">name</a>, <a href="#unitte
 ## analysistest.make
 
 <pre>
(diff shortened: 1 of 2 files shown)


UebelAndre:
@tetromino Friendly ping 😅

UebelAndre:
@tetromino One more friendly ping 😄 It'd be nice to have this in the next release

tetromino (approved):
Seems reasonable. Sorry for overlooking the PR!

tetromino:
On second thought - we made a mistake here. New parameter should go to the end of the list, otherwise we break users who use `analysistes.make` with positional parameters and expect parameter 2 to be the `expect_failure` bool. Quick fix PR incoming, please review.

PR #347, opened 2022-01-28
Add support for 'loading' unit tests, which evaluate a LOADING phase.

kkress:
This is a relatively simple addition to unittest that statically creates
rules that either explicitly fail or not depending on if the test case is valid during LOADING phase of bazel.  The test conditions are evaluated entirely in loading phase, but if we want an actual test to fail rather than just `fail()` killing the build, we can use this to assert state and report test failures.

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -564,6 +564,78 @@ def _target_under_test(env):
             fail("test rule does not have a target_under_test")
     return result
 
+def _loading_test_impl(ctx):
+    tc = ctx.toolchains[TOOLCHAIN_TYPE].unittest_toolchain_info
+    content = tc.success_templ
+    if ctx.attr.failure_message:
+        content = tc.failure_templ % ctx.attr.failure_message
+
+    testbin = ctx.actions.declare_file("loading_test_" + ctx.label.name + tc.file_ext)
+    ctx.actions.write(
+        output = testbin,
+        content = content,
+        is_executable = True,
+    )
+    return [DefaultInfo(executable = testbin)]
+
+_loading_test = rule(
+    implementation = _loading_test_impl,
+    attrs = {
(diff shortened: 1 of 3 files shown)


lib/unittest.bzl
```diff
@@ -564,6 +564,78 @@ def _target_under_test(env):
...
+def _loading_suite(env):
+    """Creates a test suite for loading phase tests.
+
+    Args:
+       env: env created with loadingtest.make
+
+    Returns:
+       None, creates test_suite
+    """
+    pass
```
tetromino, line 614: This function is no-op. Was it left over from a previous iteration of the PR?

lib/unittest.bzl
```diff
@@ -590,3 +662,9 @@ analysistest = struct(
     target_bin_dir_path = _target_bin_dir_path,
     target_under_test = _target_under_test,
 )
+
+loadingtest = struct(
+    make = _loading_make,
+    asserts = _loading_asserts,
```
tetromino, line 668: Nit: please rename to "assert", not "asserts" for consistency with the `asserts` struct defined above (in other words, we want `assert` to be a method and `asserts` to be a struct/namespace).

tetromino (changes requested):

lib/unittest.bzl
kkress: I can't do this.  assert is a keyword in starlark and fails. For now I've renamed to 'equals' to mirror the function in 'asserts'

I'm open to other suggestions.

kkress:
Failures appear to be unrelated to my changes.

tetromino:
For the failures - see https://github.com/bazelbuild/bazel-skylib/pull/349 (new buildifier release lints for more things). Maybe you can review?

tetromino (approved):
Thanks!

PR #351, opened 2022-02-10
Unbreak backwards compatibility of analysistest.make

tetromino:
Parameter 2 should be `expect_failure`. Let's not break users of positional parameters
unnecessarily.

Followup to https://github.com/bazelbuild/bazel-skylib/pull/343


UebelAndre:
@tetromino I don't know if it'd help but I opened https://github.com/bazelbuild/bazel-skylib/pull/352 so you could be the reviewer 😅

tetromino:
Closed in favor of #352

PR #352, opened 2022-02-10
Fixed positional args for unittest targets

UebelAndre:
Same change as https://github.com/bazelbuild/bazel-skylib/pull/351


tetromino (approved):
Thanks!

UebelAndre:
Sorry about the delay here. I normally just let buildifier fix things for me but noticed it fixed too much 😄 

I opened https://github.com/bazelbuild/bazel-skylib/pull/353 to make that easier for folks

PR #353, opened 2022-02-10
Applied auto-fixes for all buildifier warnings

UebelAndre:
This also enables all warnings in CI.


tetromino:
Hmm, not sure if agree. Major repos I am familiar with enable only default linter warnings, not all. What are examples of projects enabling all linter warnings?

UebelAndre:
> Hmm, not sure if agree. Major repos I am familiar with enable only default linter warnings, not all. What are examples of projects enabling all linter warnings?

I know `rules_python` and `rules_rust` both enable all lints, not sure if those are "major" though. Why wouldn't projects want to enable all warnings?

PR #358, opened 2022-03-09
Add allow_empty to target which has an empty glob

Alexander-Scott:


tetromino:
If I understand correctly, the problem you are trying to solve is that the //rules:bins filegroup is empty.

Thank you for bringing the problem to our attention via this PR, but I would prefer to solve it via https://github.com/bazelbuild/bazel-skylib/pull/359 - which is to say, instead of papering over the issue via `allow_empty`, we should simply remove the empty filegroup (and all references to it).

Issue #344, opened 2022-01-19
`diff_test::failure_message` can be used to execute sub-shells


UebelAndre:
Adding back-ticks to fail messages creates a sub-shell that leads to odd and/or undesirable behavior
https://github.com/bazelbuild/rules_python/commit/7740b22d0bae942af0797967f2617daa19834cb3#r64135831

The use of [diff_test::failure_message](https://github.com/bazelbuild/bazel-skylib/blob/1.1.1/rules/diff_test.bzl#L127) should not cause any additional commands to be executed.

tetromino:
That's rather terrible. Patches welcome, or I'll look at it when I'm back in office.

tetromino:
(The solution, of course, is to properly shell-quote `failure_message` before substituting into a bash script.)

PR #363, opened 2022-03-31
Add shell. escape_for_bat for escaping strings in Windows .bat scripts

tetromino:
In particular, we want this to be able to fix https://github.com/bazelbuild/bazel-skylib/issues/344 for both Unix and Windows.

### DIFF
--- a/lib/shell.bzl
+++ b/lib/shell.bzl
@@ -67,7 +67,8 @@ def _escape_for_bat(s, delayed_expansion = False, escape_quotes = False):
         the MSVCRT command line parser, but not by older commands like `echo`.
 
     Returns:
-      An escaped version of the string that can be passed to a command in cmd.exe.
+      An escaped version of the string that can be used as part of a command
+      line in a Windows batch file.
     """
     escaped = ""
     in_quote = False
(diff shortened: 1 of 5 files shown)


lib/shell.bzl
```diff
@@ -48,7 +48,57 @@ def _quote(s):
...
+        elif delayed_expansion and c == "!":
+            # Must be escaped twice for two expansions
+            escaped += "^^"
+
+        if c == '"':
+            if escape_quotes:
+                escaped += "\\"
+            in_quote = not in_quote
+
+        if in_quote and c == "\n":
```
comius, line 90: Should this branch happen also when `escape_quotes == True`? If I'm reading this correctly, unterminated `"` will have the newlines deleted.

lib/shell.bzl
```diff
@@ -48,7 +48,57 @@ def _quote(s):
...
+            # the only reasonable thing we can do is skip the newline
+            continue
+        if not in_quote and c in caret_escapable:
+            escaped += "^"
```
comius, line 94: Should this branch happen also when `escape_quotes == True`? This means quotes are escaped, however `caret_escapable` characters are not between those quotes.

comius (changes requested):

PR #364, opened 2022-03-31
Properly shell-quote diff_test's failure_message in bash (and document the failure_message attribute)

tetromino:
Addresses #344 for Unix; #363 is needed as prerequisite for the corresponding Windows fix.


tetromino:
Judging by CI results, on Windows, we have to both shell-escape and cmd-escape :/

Let me do some research for the right way to do this.

tetromino:
I had misdiagnosed the cause of the Windows CI failure - it came from lack of escapes in the Windows/.bat branch of _diff_test_impl, not due to any problem with shell-escaping in bash.

PR should be ok as is to at least fix escaping in Unix-like systems.

brandjon (approved):

PR #369, opened 2022-05-23
Fix doc error for analystest.begin

daivinhtran:


lib/unittest.bzl
```diff
@@ -364,6 +364,25 @@ def _begin(ctx):
     """
     return struct(ctx = ctx, failures = [])
 
+def _begin_analysis_test(ctx):
```
daivinhtran, line 367: Even though `_begin_analysis_test` is the same as `_begin`, its doc should be different from the documentation standpoint to draw a clear distinction between the two kinds of test.

tetromino: Agreed.

tetromino (approved):
Thanks for the fix. Looks good - I just needed to fix one typo.

PR #370, opened 2022-05-23
Update unittest_doc.md with analysistest and loadingtest modules

daivinhtran:


docs/unittest_doc.md
```diff
@@ -2,9 +2,11 @@
 
 Unit testing support.
 
-Unlike most Skylib files, this exports two modules: `unittest` which contains
-functions to declare and define unit tests, and `asserts` which contains the
-assertions used to within tests.
+Unlike most Skylib files, this exports four modules:
+* `unittest` which contains functions to declare and define unit tests
```
daivinhtran, line 6: Based on these modules and their use cases, it's probably a good idea to write a more complete documentation to draw a clear borderline between different kinds of tests. It seems that `unittest` supports a kind of tests different from `loadingtest` and `analysistest` even though they are all listed under "unittest_doc.md`. The doc's header says "unit testing support" but it is more than just unit testing if we refer "unit testing" as what's in `unittest`.

Should we rethink the terminologies used here? Should we reword  "unit testing support" to just "testing support" or maybe "rule testing support"?

tetromino:
@t-rad679 - good point. I took the liberty to expand the explanations.

tetromino (approved):

PR #373, opened 2022-06-10
`native_binary`: use target name as filename by default

psigen:
This changes the `out` attribute in native binary to default to the target name
if unspecified, which is convenient because it can be omitted when renaming targets
that are linux executables and frequently share the target name.

:warning: This is just a proposal for feedback.  If this looks interesting, I will clean up the change and can apply it more consistently over more of the `native_*` rules as suggested.


tetromino (changes requested):
The default is unsuitable for Windows, where executable files must have an extension. You'd need to make the rule hidden (e.g. _native_binary) and wrap it in a macro, something along the lines of
```
 def native_binary(name, src, out = None, data = None, **kwargs):
    """Move rule docs here"""
    if out == None:
        out = select({
            "@bazel_tools//src/conditions:windows": "%s.exe" % name,
            "//conditions:default": name,
        })
    _native_binary(
        name = name,
        src = src,
        out = out,
        data = data,
        **kwargs
    )
```

PR #380, opened 2022-07-23
Update README.md

BoleynSu:
If we will remove new_sets, then we should deprecate it instead of sets.


comius:
I think the docs are correct and `sets` are the ones deprecated.

BoleynSu:
Then we should fix the comments in the source file?

BoleynSu:
See https://github.com/bazelbuild/bazel-skylib/blob/main/lib/old_sets.bzl

comius (approved):

comius:
@tetromino it makes sense to point users to the `sets.bzl`, right? And get over the `new_` and `old_`, fiasco.

README.md
```diff
@@ -47,8 +47,8 @@ s = shell.quote(p)
 * [partial](docs/partial_doc.md)
 * [paths](docs/paths_doc.md)
 * [selects](docs/selects_doc.md)
-* [sets](lib/sets.bzl) - _deprecated_, use `new_sets`
-* [new_sets](docs/new_sets_doc.md)
+* [sets](lib/sets.bzl)
+* [new_sets](docs/new_sets_doc.md) - _deprecated_, use `sets`
```
comius, line 51: I think this should be
`[sets](docs/sets_doc.md)`

removing link to `new_sets` altogether and renaming the md file.

PR #381, opened 2022-08-09
Add `lib/compatibility.bzl`

philsc:
This patch adds new helper macros for dealing with incompatible target
skipping. Specifically, the macros help make the `target_compatible_with`
more readable and easier to understand.

Without these helpers you are left to write verbose `select()` statements.

For example, you might write the following to express a target that is
incompatible with QNX and bare-bones systems.
```python
cc_library(
    name = "lib",
    srcs = ["lib.cc"],
    target_compatible_with = select({
        "@platforms//os:qnx": "@platforms//:incompatible",
        "@platforms//os:none": "@platforms//:incompatible",
        "//conditions:default": [],
    }),
)
```

WIth one of the helpers from this patch you can instead rewrite this as:
```python
cc_library(
    name = "lib",
    srcs = ["lib.cc"],
    target_compatible_with = compatibility.none_of([
        "@platforms//os:qnx",
        "@platforms//os:none",
    ]),
)
```

That is now easier to understand. Especially for folks new to
incompatible target skipping.

This patch is based on @trybka's trybka/bazel-skylib#1. The
implementation is his. I cleaned up the tests and added documentation.


lib/compatibility.bzl
```diff
@@ -0,0 +1,171 @@
...
+      A string which is a valid target name.
+    """
+    targets_name = joiner.join([s.split(":")[1] for s in targets])
+    name = targets_name.replace("//", "").replace(":", "_").replace("/", "_").replace(".", "_")
+    return name
+
+def _maybe_make_unique_incompatible_value(name):
+    """Creates a `native.constraint_value` which is "incompatible."
+
+    When composing selects which could all resolve to "incompatible" we need distinct labels.
```
fmeum, line 33: Just out of curiosity: Is there a particular reason for why `target_compatible_with` can't deduplicate it's value before processing it? That won't help simplify this for existing Bazel versions though.

lib/compatibility.bzl
```diff
@@ -0,0 +1,171 @@
...
+    Removes/replaces characters which are not valid as target names.
+
+    Args:
+      target_list: A list of target names.
+      joiner: An optional string to use for joining the list.
+
+    Returns:
+      A string which is a valid target name.
+    """
+    targets_name = joiner.join([s.split(":")[1] for s in targets])
```
fmeum, line 26: Not all target names contain `:`, the two exceptions I know of are short-hands (`//pkg/main` for `//pkg/main:main`) and same-package targets (`some_name` for `:some_name`).

lib/compatibility.bzl
```diff
@@ -0,0 +1,171 @@
...
+    name = targets_name.replace("//", "").replace(":", "_").replace("/", "_").replace(".", "_")
+    return name
```
fmeum, line 28: You could increase the chance that this name is unique even further by appending something like `str(hash(tuple(targets)))`.

trybka:
Oh hey -- thanks for picking this up!

lib/compatibility.bzl
philsc: Good point. I will fix it.

lib/compatibility.bzl
philsc: I suspect that the reason is that there's some fairly generic logic somewhere that complains about duplicate labels in `label_list` attributes. E.g. I remember that adding duplicate deps to `py_library` also throws a bazel error. In other words, I suspect that this behaviour cannot be changed without affecting the same logic for all other attributes. Unless we hard-code something for `target_compatible_with`.

lib/compatibility.bzl
philsc: Neat idea. I will think about it. Need to run through the logic and clarify the requirements in my head again.

lib/compatibility.bzl
fmeum: True. It's pretty easy to hit this error when programmatically adding dependencies from a macro.

@gregestren Can you perhaps shed some light on why Bazel errors out rather than stably deduping the label list?

lib/compatibility.bzl
philsc: Done.

lib/compatibility.bzl
philsc: Done

philsc:
If anyone has an idea why bazel is segfaulting on shutdown in one of the Windows tests, I'd appreciate it.

philsc:
Any of the 5 code owners interested in taking a look at this PR?

philsc:
@brandjon , @tetromino , @comius , @hvadehra , @c-mita , any thoughts? I am not sure what to make of the complete lack of feedback here.

docs/compatibility_doc.md
```diff
@@ -0,0 +1,145 @@
...
+load("@bazel_skylib//lib:compatibility.bzl", "compatibility")
+```
+
+See the [Platform docs](https://bazel.build/docs/platforms#skipping-incompatible-targets) for
+more information.
+
+
+<a id="#compatibility.all_of"></a>
+
+## compatibility.all_of
```
comius, line 16: Is this function really necessary? 
I think the example should work with `target_compatible_with =[":dbg", "@platforms//os:linux"]`, that is by default conjunction is used.

lib/compatibility.bzl
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns:
+      A native `select()` which is "incompatible" unless all settings are true.
+    """
+    group_name = _get_name_from_target_list(settings, joiner = " and ")
+    compat_name = " compatible with all of " + group_name
+    _maybe_make_unique_incompatible_value(compat_name)
+
+    # all_of can only be accomplished with a config_setting_group.match_all.
+    if not native.existing_rule(group_name):
```
comius, line 170: We'd like to get rid of `native.existing_rule` and we don't really support "anonymous" target definitions. Are there other ways how to implement this without using it? 

Maybe copying `selects.config_setting_group` pattern, that is add a target for `all_of` and then refer to it in the target_compatible_with.

Another alternative is to construct and return `select`s, without creating a target. I know this is possible for `match_any(a,b)` (by using `select{a:a,b:b}`) and it seems to be possible for `match_all` (just returning a list should do it, if I'm not mistaken). It might even work with some types of compositions.

lib/compatibility.bzl
```diff
@@ -0,0 +1,185 @@
...
+    This will create a `constraint_value` with the given name, if it does not already exist.
+
+    Args:
+      name: A target name to check and use.
+    """
+    if not native.existing_rule(name):
+        native.constraint_value(
+            name = name,
+            constraint_setting = "@platforms//:incompatible_setting",
+        )
```
comius, line 57: There already is a `constraint_value`: `@platforms//:incompatible`

lib/compatibility.bzl
```diff
@@ -0,0 +1,185 @@
...
+    """
+    compat_name = " compatible with any of " + _get_name_from_target_list(settings)
+    _maybe_make_unique_incompatible_value(compat_name)
+
+    return selects.with_or({
+        tuple(settings): [],
+        "//conditions:default": [":" + compat_name],
+    })
+
+def _all_of(settings):
```
comius, line 129: Why not `*settings`? That might be more usable.

comius (changes requested):

comius:
>

comius:
Ups, wrong button.

> any thoughts? I am not sure what to make of the complete lack of feedback here.

Too many reviewers - each one of them has time to review, but all of them together don't have it :)

docs/compatibility_doc.md
trybka: This supports any `config_setting`, not just constraints. I think `target_compatible_with` only works with just a list of constraints by default.

lib/compatibility.bzl
philsc: I think we could solve this by chaining `select()` statements, but only if we pre-define a pre-determined number of incompatible `constraint_value` targets. E.g. if we wanted to support up to 50 settings in `all_of`, we could do this:
```python
# lib/compatibility/BUILD
[constraint_setting(
    name = "all_of_" + str(i),
    constraint_value = "@platforms//:incompatible_setting",
) for i in range(50)]
```
Then `all_of` could be implemented roughly like so:
```python
def all_of(*settings):
    result = []
    for i, setting in enumerate(settings):
        result += select({
            setting: [],
            "//conditions:default": "@bazel_skylib//lib/compatibility:all_of_" + str(i),
        })
    return result
```
So if _any_ of the `select`s fail, then the target would be incompatible.

Am I thinking about this correctly?

philsc: I took the liberty of implementing this. What do you think?

lib/compatibility.bzl
philsc: I ended up implementing the approach I thought of in the above comment. Please let me know what you think. Happy to add more comments etc. if you approve of the approach.

philsc:
@comius , any thoughts?

comius:
> @comius , any thoughts?

I did a quick pass. Looks nice and looks ok, thanks. I'll do another more thorough pass and let you know.

lib/compatibility/BUILD
```diff
@@ -0,0 +1,32 @@
...
+
+constraint_value(
+    name = "none_of",
+    constraint_setting = "@platforms//:incompatible_setting",
+)
+
+constraint_value(
+    name = "any_of",
+    constraint_setting = "@platforms//:incompatible_setting",
+)
```
comius, line 27: The name `any_of` is confusing. Reading the code for `_any_of`, I had no idea it works until I looked up how `any_of` is actually implemented.

I can see now, you named constraints by the location where they are used, but that's confusing. 

Is there a reason you're defining new `constraint_values` and not using directly `@platforms//:incompatible`, which is a constraint value already?

If not, I'd suggest you just remove all `constraint_value`-s and use the predefined one in the functions.

lib/compatibility.bzl
```diff
@@ -0,0 +1,130 @@
...
+
+    Args:
+      *settings: The `config_setting` or `constraint_value` targets.
+
+    Returns:
+      A native `select()` which maps any of the settings to the incompatible target.
+    """
+    return selects.with_or({
+        "//conditions:default": [],
+        tuple(settings): ["@bazel_skylib//lib/compatibility:none_of"],
```
comius, line 44: Nit: I would find this slight more "understandable" if you put "//conditions:default" at the end. If I understand the semantics of select correctly - it's evaluated top to bottom, with default always being last.

lib/compatibility.bzl
```diff
@@ -0,0 +1,130 @@
...
+      A native series of `select()`s. The result is "incompatible" unless all settings are true.
+    """
+    if len(settings) > MAX_NUM_ALL_OF_SETTINGS:
+        fail("Cannot support more than {} arguments. Use selects.config_setting_group() instead.".format(MAX_NUM_ALL_OF_SETTINGS))
+
+    result = []
+    for i, setting in enumerate(settings):
+        result += select({
+            setting: [],
+            "//conditions:default": ["@bazel_skylib//lib/compatibility:all_of_{}".format(i)],
```
comius, line 122: What happens if you replace `"@bazel_skylib//lib/compatibility:all_of_{}".format(i)` with `"@platforms//:incompatible"` in this line?

It if works, it's much simpler.

tests/compatibility_test.sh
```diff
@@ -0,0 +1,324 @@
...
+  cat >> target_skipping/BUILD <<EOF
+sh_test(
+    name = "pass_on_foo1_or_foo2_but_not_bar1",
+    srcs = [":pass.sh"],
+    target_compatible_with = compatibility.any_of(
+        ":foo1",
+        ":foo2",
+    ) + compatibility.none_of(
+        ":bar1",
+    ),
```
comius, line 294: So `any_of(foo1)+none_of(foo1)` is always false, right?
What about `all+none`, `any+all`, `any+any`, `all+all`, `none+none`, they all work right?

comius (changes requested):

lib/compatibility/BUILD
philsc: This goes back to the earlier discussion: https://github.com/bazelbuild/bazel-skylib/pull/381#discussion_r954544755

The short version is that bazel is not happy about duplicated labels in a label list.

lib/compatibility.bzl
philsc: Same issue as https://github.com/bazelbuild/bazel-skylib/pull/381#discussion_r954544755

lib/compatibility.bzl
philsc: That's fair. I will fix that in the next iteration.

tests/compatibility_test.sh
philsc: > So any_of(foo1)+none_of(foo1) is always false, right?

I have not added a test to explicitly test this, but I don't see any way that it would ever evaluate to true.

> What about all+none, any+all, any+any, all+all, none+none, they all work right?

Theoretically yes. The challenge at the moment is that because of bazel's dislike of duplicated labels in label lists, we can't do, say, any+any, all+all, or none+none. I think the only way to make that work is to either:
1. Figure out how to make bazel not complain about duplicate labels in a label list, or
2. Accept a `name`-like parameter for these helpers that will create new `constraint_value` targets for that invocation.

lib/compatibility/BUILD
philsc: To elaborate a little bit: In today's bazel, you cannot freely compose everything with `@platforms//:incompatible` because you risk getting the label into the final label list more than once. Bazel will then error out.

comius: Ack. Sorry, I forgot about that. Can you then rename to something like `incompatible_in_any_of`? And please "sprinkle" some comments about this problem :)

lib/compatibility.bzl
comius: There's one alternative, which I considered worse until now. How about:
```
result = []
for setting in enumerate(settings):
   result += select({
      setting: result,
       "//conditions:default": ["@bazel_skylib//lib/compatibility:incompatible_in_all_of"]
```

I think this should evaluate to a single label when a condition is not met.

First line could also be `result = settings`. That might work better in `cquery`-es.

tests/compatibility_test.sh
comius: I've got a feeling there is something fishy going on, and I'm not able to put my finger on it. There is negation without universal set and `+` seems to mean the same as conjunction. The feeling is you're not using the right set of primitives. But I'm failing to figure out what those should be.

For example using `+` for conjunction, `not` for single negation and `any` for disjunction. This way you should be able to write any boolean expression as something like `= any(a,b,not(c)) + any(c,d,not(e))`. But then `a+b` only works for constraints. Should you also have `as_constraints`, that wraps a config value.

I can't really tell what's more usable. So far I've seen the most places need just conjunction of constraints and some more rare places a disjunction of constraints.

comius: Thinking out loud: maybe opposite approach is the solution: a single function `match([a,b],c,[d,e], noneof=[f,g])` which says `(a and b) and c and (d or e) and (not f) and (not g)`.

With current situation about composition - I would be more happy to make no guarantees about it, because they seem to be quite "flaky". Maybe even better, breaking composition on purpose, by using that single `incompatible` constraint value.

philsc: Filed bazelbuild/bazel#16274 to see what the usual suspects think about it.

comius: One more thought. What if we allow config_values in the list? I think this is doable in Bazel. @katre?

This could simplify a lot what's going on here, specially `all_of` would become unnecessary. I believe the initial design decision was not to mix constraints and configuration, but we could revisit that.

philsc: One of the pieces of wisdom that @gregestren imparted on me early on is that if something can reasonably simply be implemented in Starlark instead of native bazel, then Starlark is the preferred option. I.e. keep things out of bazel if there's a reasonable approach to doing it in Starlark. I have not yet found anything that has convinced me otherwise. In this case, I'm not convinced that we have to make bazel's implementation more complicated just to make the Starlark implementation here simpler.

Of course I will let you all be the judge of whether or not using a series of `select()` statements is a reasonable approach to avoid making bazel more complicated.

katre: What is meant by `config_value`? Is `config_setting` meant?

philsc: I'm assuming so. I myself can't for the life of me keep them straight. The fact that `select()` accepts config_**setting**s and constraint_**value**s makes me accidentally mix the two very frequently.

katre: Using `config_setting` in `target_compatible_with` is confusing: what does it mean if a target is `target_compatible_with` a setting that checks whether flag `--compilation_mode` is set to `OPT`? Either we define semantics for the full set of `config_setting` behaviors, or we're just using it to group `constraint_values`, and I'd argue that isn't worthwhile.

I agree that the similarity and differences between `config_setting` and `constraint_value` is unfortunate, but even several years later I haven't been able to come up with better names.

philsc: I agree that with bare bazel, there aren't great examples for `target_compatible_with` and `config_setting`.

I suspect most of the interesting use cases come from user-defined `config_setting` targets. For example, we internally use `target_compatible_with` to mark things as only compatible with `clang` vs. `gcc`. That doesn't actually have anything to do with target platforms, but it's a good mechanism. We also use this to mark code as compatible with certain versions of Python. We don't encode the Python version in the target platform since everything is hermetic. In short, we use it to make assertions about code compatibility with toolchains. The only way that is possible today is with `target_compatible_with` and `config_setting`.

Regardless, I do technically like your `match()` idea. I just need to figure out how to make it more understandable when you don't know what the function does. That's really the only reason I haven't switched to it yet.

In the meantime, I've switched the implementation here to use `@platforms//:incompatible` for everything. The tests all pass when run against bazel HEAD.

I will keep thinking about how to make the `match()` idea work well.

philsc: @comius , I spent some time implementing your `match` idea, but I'm not convinced that it'll be ergonomic. Starting with this:
```
target_compatible_with = compatibility.match([":a", ":b"]),  # This means a _or_ b.
target_compatible_with = compatibility.match(":a", ":b"),  # This means a _and_ b.
```
makes me think the difference is too subtle. Making it more verbose:
```
target_compatible_with = compatibility.match_all(  # This means a _or_ b.
    compatibility.or_group(":a", ":b"),
),
target_compatible_with = compatibility.match_all(":a", ":b"),  # This means a _and_ b.
```
makes it easier to understand But then we're not really saving the user much over doing:
```
target_compatible_with = compatibility.any_of(":a", ":b"),  # This means a _or_ b.
target_compatible_with = compatibility.all_of(":a", ":b"),  # This means a _and_ b.
```

With bazel HEAD, it's actually pretty easy now to compose the `match([a,b],c,[d,e], noneof=[f,g])` example from earlier.
```
target_compatible_with = compatibility.any_of(a, b) + \
                         compatibility.all_of(c) + \
                         compatibility.any_of(d, e) + \
                         compatibility.none_of(f, g),
```

WDYT about the current state of things?

aiuto: I also think different behavior for list of conditions vs. a single set is far too subtle. People will get that wrong.
any_of and all_of are excellent names.

lib/compatibility.bzl
gregestren: @philsc and I followed up on this on another issue/PR (which was that?). I assume the historic logic is to encourage tidy BUILD curation. I ran this by the devs most concerned about API consistency and didn't hear strong opinion on keeping it this way.

tests/compatibility_test.sh
philsc: @comius , any thoughts?

I can disable the tests for pre-6.0 versions of bazel. But that would mean not running tests at all for now.

philsc: @comius , WDYT?

philsc: @comius , ping

PR #383, opened 2022-08-16
Fixing typo: allocate slice with the correct size

### LINKED ISSUE #382: Question: redundant slice element in gazelle Import() function?
While studying the source code, I came across this line:
https://github.com/bazelbuild/bazel-skylib/blob/872e9b06e18ae8ba2897cb65b9aaa172aa6279f3/gazelle/bzl/gazelle.go#L114

Later on, `imports` will be appended by as many as `len(srcs)` elements:
https://github.com/bazelbuild/bazel-skylib/blob/872e9b06e18ae8ba2897cb65b9aaa172aa6279f3/gazelle/bzl/gazelle.go#L116-L126

At the end, `len(imports)` and `cap(imports)` both become `2*len(srcs)`, which seems odd to me because at first it was only allocated for `len(srcs)` elements.

Does it make sense to change
`imports := make([]resolve.ImportSpec, len(srcs))`
to
`imports := make([]resolve.ImportSpec, 0, len(srcs))`?

Note that I am very new to golang, so chances are high that I could be wrong.

hchsiao:
Attempting to fix #382

### DIFF
--- a/gazelle/bzl/gazelle.go
+++ b/gazelle/bzl/gazelle.go
@@ -111,7 +111,7 @@ func (*bzlLibraryLang) Fix(c *config.Config, f *rule.File) {}
 // returned, including an empty slice, the rule will be indexed.
 func (b *bzlLibraryLang) Imports(c *config.Config, r *rule.Rule, f *rule.File) []resolve.ImportSpec {
 	srcs := r.AttrStrings("srcs")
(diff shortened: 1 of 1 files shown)


adonovan (approved):

adonovan:
attn: @tetromino

tetromino (approved):
Thanks!

PR #385, opened 2022-08-29
feat: bzlmod setup

nickgooding:
Creates MODULE.bazel and WORKSPACE.bzlmod files to enable bazel-skylib
to work with the `--experimental_enable_bzlmod` flag.

rules_go has been updated as the previously used version is not
available in the BCR.

stardoc has been updated due to a strange issue with Bzlmod enabled that
caused the `diff_test` and `unittest` docs to be generated without any
contents that was magically fixed upon updating.

bazelbuild/bazel-central-registry#124

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,6 +10,10 @@ register_toolchains(
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
+
+### INTERNAL ONLY - lines after this are not included in the release packaging.
+
+# Gazelle extension is experimental
 bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.33.0")
 bazel_dep(name = "gazelle", repo_name = "bazel_gazelle", version = "0.26.0")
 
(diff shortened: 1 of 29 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,26 @@
+module(
+    name = "bazel_skylib",
+    compatibility_level = 1,
+    version = "1.2.1",
+)
+
+register_toolchains(
+    "//toolchains/unittest:cmd_toolchain",
+    "//toolchains/unittest:bash_toolchain",
+)
```
tetromino, line 10: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/bazel_skylib/1.2.1/MODULE.bazel instead uses 

```
module(
    name = "bazel_skylib",
    compatibility_level = 1,
    toolchains_to_register = [
        "@bazel_skylib//toolchains/unittest:cmd_toolchain",
        "@bazel_skylib//toolchains/unittest:bash_toolchain",
    ],
    version = "1.2.1",
)
```

AFAICT, both `toolchains_to_register` and `register_toolchains` allow bazel_skylib's own tests to pass. Which one is preferred?

tetromino (changes requested):
Thanks for doing this!

The only issue is that we'd need to adapt MODULE.bazel to work with bazel_skylib's distribution tarball: we strip out all build-only dependencies and the gazelle extension (which is still experimental) to build the release tarball - see distribution/BUILD.

I think the best way to do that would be to cut everything below the platforms dep out of MODULE.bazel and put the result in the //:distribution filegroup.

I could make those changes myself, fwiw.

tetromino (approved):
Added a magic comment so that MODULE.bazel can be split into distro-able and non-distroable pieces via sed

PR #387, opened 2022-08-30
Add missing docs and doc links; update bazel.build links

tetromino:


docs/select_file_doc.md
```diff
@@ -0,0 +1,28 @@
+<!-- Generated with Stardoc: http://skydoc.bazel.build -->
+
+
+select_file() build rule implementation.
+
+Selects a single file from the outputs of some target by given relative path.
```
comius, line 6: nit: here and below "some target" sounds weird, but I'm not an expert in English. "a target" or "some targets" would sound better to me.

comius (approved):

PR #392, opened 2022-09-01
Allow sandboxing for copy_* and fix copy_directory tests

tetromino:
After some thought, I have to say that forcing a local strategy for
copy_file/copy_directory is inappropriate. The point of a sandbox is to
catch hermeticity bugs; disabling the sandbox may be useful for
performance, but it's up to the user to do it if they trust us - and
they can do it via flag. The default should be paranoia and safety.

And on the subject of strategies - using a genrule to create an empty
directory fails in environments where genrules run remote by default
(and thus, copy_directory tests fail). We could, of course, set local=1,
but that disables the sandbox and causes scary warnings. Instead, add a
proper empty_directory rule to test with.

### DIFF
--- a/tests/copy_directory/empty_directory.bzl
+++ b/tests/copy_directory/empty_directory.bzl
@@ -1,0 +1,34 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Creates an empty directory."""
+
+def _empty_directory_impl(ctx):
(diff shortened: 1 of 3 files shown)


tests/copy_directory/empty_directory.bzl
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Creates an empty directory."""
+
+def _empty_directory_impl(ctx):
+    out = ctx.actions.declare_directory(ctx.attr.name)
+    ctx.actions.run_shell(
+        outputs = [out],
+        command = "mkdir -p $@",
```
comius, line 21: "-p" might not work on windows (https://stackoverflow.com/questions/905226/what-is-equivalent-to-linux-mkdir-p-in-windows)

However, the change already makes it work in more cases on linux (when name = "a/b").

I wonder if such PRs should be our top priority.

comius (approved):

tests/copy_directory/empty_directory.bzl
comius: PS this might work with just `ctx.actions.do_nothing`. I know Bazel creates some directory for the action, not sure if output directory is among them.

tetromino: It's a test-only rule, so I think it's reasonable to require bash on windows for it, as many of our tests already do.

re top priority - it's a step towards making it possible to update Google's internal skylib to 1.3.0.

re `ctx.actions.do_nothing` - unfortunately, we require an action that generates `out`.

PR #394, opened 2022-09-05
Make settings error message more friendly

Wyverald:
By stripping leading '@'s from labels in the main repo. So it talks about the flag '//foo:bar' instead of the flag '@//foo:bar'.


Wyverald:
cc @meteorcloudy

Wyverald:
This is needed to flip https://github.com/bazelbuild/bazel/issues/16196

tetromino (approved):
Looks good!

Issue #386, opened 2022-08-30
Keep version in sync between MODULE.bazel and version.bzl


tetromino:
Reminder to myself: after merging https://github.com/bazelbuild/bazel-skylib/pull/385, we will need some sort of a diff test to keep the module version and the `version.bzl` version in sync; neither file can load the other - MODULE.bazel prohibits `load` statements.

tetromino:
The natural way to do so would be to extract `version` from MODULE.bazel via buildozer:

```
cat MODULE.bazel | buildozer 'print version' '-:bazel_skylib'
```

Unfortunately, this means we need to depend on buildozer, and while this is easy enough to do via `WORKSPACE`, it fails with `--experimental_enable_bzlmod`:

```
error loading package '@gazelle~0.26.0~go_deps~com_github_bazelbuild_buildtools//api_proto': Unable to find package for @[unknown repo 'rules_proto' requested from @gazelle~0.26.0~go_deps~com_github_bazelbuild_buildtools]//proto:defs.bzl: The repository '@[unknown repo 'rules_proto' requested from @gazelle~0.26.0~go_deps~com_github_bazelbuild_buildtools]' could not be resolved: Repository '@rules_proto' is not visible from repository '@gazelle~0.26.0~go_deps~com_github_bazelbuild_buildtools'. and referenced by '@gazelle~0.26.0~go_deps~com_github_bazelbuild_buildtools//edit:go_default_library'
```

I believe that I'm hitting https://github.com/bazelbuild/rules_go/issues/3265

Alternatively, we could avoid the problem trivially if there were an official buildtools module in the central registery - but it doesn't exist yet.

Current work in progress is sitting at https://github.com/tetromino/bazel-skylib/tree/version_consistency

fmeum:
@tetromino I just submitted a PR that should add support for `go_proto_library` to rules_go with Bzlmod: https://github.com/bazelbuild/rules_go/pull/3283. I manually verified that it can build buildozer.

fmeum:
rules_go 0.34.0 is now available in the register and should support `go_proto_library`.

tetromino:
> rules_go 0.34.0 is now available in the register and should support `go_proto_library`.

0.34.0 was released in July. I think your fix is in 0.35.0, which isn't in the register yet :)

fmeum:
[...]

PR #396, opened 2022-09-08
Add a version consistency check

tetromino:
Extract the version from version.bzl using sed.

For MODULE.bazel, we need to use buildifier to avoid confusion between skylib's version and our dependencies' versions.


tetromino:
NOTE: this is not ready to merge yet - broken with `--experimental_enable_bzlmod`. See https://github.com/bazelbuild/bazel-skylib/issues/386#issuecomment-1241174956

PR #397, opened 2022-09-25
Bump platforms to latest version in registry

hanneskaeufler:
Just to be more up to date when eventually `platforms` exposes it's own MODULE.bazel file.


aiuto:
Do we really need to do these kind of updates?
We're getting a lot of churn on deps of core rules, but there is no compelling need for many of them.
 It feels like we should adopt some policy that coordinates the need for these with patch releases.

Issue #250, opened 2020-05-24
Automatic generation of `skylark_library` targets in `BUILD` files


achew22:
Hey @aiuto and @laurentlb (who appear to be OWNERS on this repo),

First, thanks so much for all you're doing in the Bazel ecosystem. I hugely appreciate your efforts. One of the things that I have been thinking about creating for a while was a [bazel-gazelle](https://github.com/bazelbuild/bazel-gazelle) plugin for `skylib` that generated `skylark_library` targets. I think this would be generally useful. In my repo's I've found it helpful to have `skylark_library` targets to feed into [`stardoc`](https://github.com/bazelbuild/stardoc) and manually creating the targets is kind of a drag.

Fortunately, Gazelle has a really nice plugin system that can be used to add generation for languages beyond golang. If I were to write a skylib plugin for Gazelle, would this repo be a good place to host it? Since the repo is packaged by `rules_pkg` I could iterate on it and not have it released as part of the distribution bundle until it was deemed to be good enough. Additionally I would be happy to update the distribution scripts to support both a gazelle version and a non-gazelle version to keep down dependencies if that's a concern.

Please let me know what you think. Thanks so much!

aiuto:
cc: @jayconrod  
I really don't know where that would belong. Do gazelle plugins go with a gazelle contrib folder or where they seem used?

Then we have the question of how to bundle libraries for the optimal stardoc generation.  One bzl_library per file is probably not wrong. If startdoc is going to do something for integrated doc sets, that could be structured at the stardoc rule level, rather than having a big library.

jayconrod:
Ideally I think Gazelle extensions should live in the same repos where the rules are defined. That way, the extension can be coupled with the rules being used. Rule authors are also in a much better position to maintain those extensions than I am.

From the not-following-my-own-advice department, the Go extension lives in the Gazelle repository (Gazelle was originally Go-only). There have been some issues with Gazelle and rules_go versions drifting apart.

achew22:
In order to push the discussion along, I decided to implement a strawman proposal. Please see #251 for that proposal. I have a few questions in there for both of you. If you would be so kind as to review and respond to them I would greatly appreciate it.

aiuto:
To be honest, I am not going to have the time to look at this for a while. Nor can I commit to approving, because I am not directly a skylib maintainer. 
@laurentlb @c-parsons  Can either of you comment on the overall concept of doing this.

achew22:
[...]

PR #400, opened 2022-10-06
Add the gazelle plugin to the distribution

shs96c:
To avoid everyone needing to take a dep on `rules_go`, we do the following:

1. Regular Bazel users need to load `bazel_skylib_gazelle_plugin_workspace` and call that, and then `bazel_skylib_gazelle_plugin_setup`

2. `bzlmod` users need do nothing, but we now include the `rules_go` dep in the `MODULE.bazel` shipped in the release. This is fine, because `bzlmod` will lazily load dependencies.


MODULE.bazel
```diff
@@ -27,6 +25,8 @@ go_deps.module(
 
 use_repo(go_deps, "com_github_bazelbuild_buildtools")
 
+### INTERNAL ONLY - lines after this are not included in the release packaging.
```
fmeum, line 28: I think that you can move the `### INTERNAL ONLY` header above the `go_deps = use_extension` line - that is only used to load buildifier for internal use within the bazel-skylib workspace.

shs96c: Done

nickgooding: The Gazelle plugin requires `@com_github_bazelbuild_buildtools`, so I don't think it will build without this.
https://github.com/bazelbuild/bazel-skylib/blob/bc112d41fd71a5abf51ff827f19a6af9ed81af43/gazelle/bzl/BUILD#L19

fmeum: Didn't think of that, thanks for pointing it out.

shs96c: Moved :)

gazelle/bzl/BUILD
```diff
@@ -51,3 +51,25 @@ gazelle(
     name = "gazelle",
     gazelle = ":gazelle-skylib",
 )
+
+# The files needed for distribution
```
fmeum, line 55: Would adding `allow_empty` to the `glob` be an alternative? It wouldnt' be error-prone in this context.

gazelle_setup.bzl
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+"""Dependency registration helpers for the gazelle plugin."""
+
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+
+def bazel_skylib_gazelle_plugin_setup():
+    go_rules_dependencies()
+    go_register_toolchains(version = "1.17.1")
```
fmeum, line 22: Have you tried calling this macro from a `WORKSPACE` that already sets up a Go toolchain somewhere else? You might have to make running this configurable via a parameter.

If possible, also bump the version to `1.18.7` as `1.17` is out of support. `1.19` is available, but requires a relatively recent version of `rules_go`.

fmeum (approved):
I think you may have removed all reviewers with write access to this repo :-)

gazelle/bzl/BUILD
shs96c: According to the docs, `allow_empty` defaults to `True`, but I've made it explicit.

gazelle_setup.bzl
shs96c: Done and done.

shs96c:
@fmeum I thought you could commit to this repo :) Oops. Will add an owner as a reviewer!

Wyverald:
Have you tried out the approach we discussed in #250 where the part using gazelle is separated out into its own module? I still think it's rather intrusive to have rules_go in basically everyone's dependency graph.

shs96c:
As I understand it, `bzlmod` will only lazily load the dependencies, so it should make any difference to the time it takes to build a project. Having `rules_go` in the dependency graph will mostly be a no-op. I don't think users will be in the habit of expecting more than one module per repo, so the user experience of this approach is "lower friction" for `bzlmod` users.

For people not using `bzlmod`, they need to actively pull in the `rules_go` dep by calling `gazelle_setup`: it's not part of the default workflow.

fmeum:
I obviously don't like saying that, but unfortunately having a transitive Bzlmod dep on `rules_go` isn't really a no-op (yet):

1. Since we are still in a phase of rapid iteration on `rules_go`'s Bzlmod interface, it emits warnings like the following when depended on:
```
DEBUG: gazelle@0.26.0/MODULE.bazel:6:6: WARNING: The bazel_gazelle Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
DEBUG: rules_go@0.33.0/MODULE.bazel:7:6: WARNING: The rules_go Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
```
We will try to stabilize this shortly after the Bazel 6 release, but I can't promise anything.

2. I am still working on making the toolchain fetching in `rules_go` lazy. Currently, it does perform some network I/O if it is depended on even if you don't build Go targets.

At least before both these issues have been fixed introducing a `rules_go` dependency on a core module such as skylib won't be a good user experience. It should be pretty simple to migrate users off of a separate module though after these limitations have been addressed.

Wyverald:
> As I understand it, `bzlmod` will only lazily load the dependencies, so it should make any difference to the time it takes to build a project.

That's true, but as Fabian also pointed out, bringing in other modules sometimes *does* fetch them eagerly, particularly when toolchains are registered in them. Besides, adding dependencies has other costs -- version resolution takes more time and includes more noise, and a "vendor mode" setup ("download everything for me so that I can work offline") will fetch much more random stuff.

> I don't think users will be in the habit of expecting more than one module per repo, [...]

If by "repo" here you mean "git repo" instead of "Bazel repo", we need to normalize the expectation of multiple modules hosted in the same git repository. That would be the only sane way to avoid an explosion of the dependency graph.

shs96c:
@Wyverald updated to be a two module, one git repo setup.

shs96c:
The build is failing with `last_green` because a target in `rules_pkg` isn't visible:

```
ERROR: /workdir/distribution/BUILD:17:8: in pkg_tar_impl rule //distribution:bazel-skylib-1.3.0: target '@rules_pkg//private:private_stamp_detect' is not visible from target '//distribution:bazel-skylib-1.3.0'. Check the visibility declaration of the former target if you think the dependency is legitimate
ERROR: /workdir/distribution/BUILD:17:8: Analysis of target '//distribution:bazel-skylib-1.3.0' failed
```

With `last_rc` and Bazel 5.3.2, this doesn't appear to be a problem.

fmeum:
@shs96c HEAD broke downstream projects, should be fixed in about an hour thanks to https://github.com/bazelbuild/bazel/commit/7dcf0c350536903a3c4dbeb514bf332b52275458.

gazelle/MODULE.bazel
```diff
@@ -0,0 +1,24 @@
+bazel_dep(name = "bazel_skylib", version = "1.3.0")
```
Wyverald, line 1: add a `module(name="skylib_gazelle", version="...")` declaration here?

MODULE.bazel
```diff
@@ -14,8 +14,8 @@ bazel_dep(name = "platforms", version = "0.0.4")
 ### INTERNAL ONLY - lines after this are not included in the release packaging.
 
 # Gazelle extension is experimental
-bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.33.0")
-bazel_dep(name = "gazelle", repo_name = "bazel_gazelle", version = "0.26.0")
+bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.35.0")
```
Wyverald, line 17: is the plan to remove these?

Wyverald (commented):
nice start!

MODULE.bazel
shs96c: I can try removing them and see what breaks....

gazelle/MODULE.bazel
shs96c: Done.

MODULE.bazel
shs96c: Removed.

WORKSPACE
```diff
@@ -69,17 +69,5 @@ register_unittest_toolchains()
 # github.com/bazelbuild/rules_go is available from io_bazel_rules_go and it
 # doesn't need to duplicatively fetch it.
 # gazelle:repository go_repository name=io_bazel_rules_go importpath=github.com/bazelbuild/rules_go
```
Wyverald, line 71: I'm not familiar with gazelle repository hints, do these need to be moved to gazelee/WORKSPACE?

Wyverald (commented):
all looks good except for the question about the gazelle hints left over in the WORKSPACE file. thanks!

Wyverald (approved):

WORKSPACE
shs96c: I've moved them, and everything still seems to build. Apparently, it's fine to move them.

Wyverald (approved):

cgrindel:
@shs96c Will we be adding instructions on how to load the `@bazel_skylib_gazelle_plugin` repository into a workspace?

PR #402, opened 2022-10-14
Stop hardcoding runfiles prefix

nickgooding:
bazelbuild/bazel@6a8ddb7 changed the prefix for runfiles from the main repo when Bzlmod is enabled. Because all uses of rlocation were hardcoded to use "bazel_skylib", tests requiring runfiles stopped working with Bzlmod enabled. This commit updates calls to rlocation to instead use the TEST_WORKSPACE env var to get the repo name.

### DIFF
--- a/tests/analysis_test_test.sh
+++ b/tests/analysis_test_test.sh
@@ -41,7 +41,7 @@ else
 fi
 # --- end runfiles.bash initialization ---
 
-source "$(rlocation bazel_skylib/tests/unittest.bash)" \
+source "$(rlocation $TEST_WORKSPACE/tests/unittest.bash)" \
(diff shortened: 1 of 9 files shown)


fmeum:
We are working on a conceptual solution to this at https://github.com/bazelbuild/bazel/issues/16124, but this is a decent workaround in the meantime.

fmeum:
@Wyverald

Wyverald (approved):
thanks!

nickgooding:
@fmeum Thanks, I did see that issue while trying to find a fix for the tests in this repo, but I decided it would be better to just get something working for now. I'm happy to circle back to this once the solution has firmed up a bit.

PR #405, opened 2022-10-27
Update rules_go to fix build with bzlmod

nickgooding:
Since bazelbuild/bazel@3f92232, rules_go v0.33.0 has been broken with bzlmod due to missing @com_github_golang_mock repository


fmeum:
This update is still welcome, but the breakage may have already been fixed by https://github.com/bazelbuild/bazel/commit/cf3f48ca49f089615417636763d753811acf717f.

nickgooding:
You're right. I didn't notice that commit when I began looking at this. 🤦

PR #411, opened 2022-11-11
Stop registering a Go SDK in the Gazelle submodule

shs96c:
I'm not sure if this is the best way to solve this problem, but there are two cases where we want to use this repo:

1/ We're developing the repo itself. We should be responsible for
   registering the Go toolchain.

2/ We're being used transitively by another module (eg.
   `contrib_rules_jvm`. We should _not_ be responsible for
   registering the Go toolchain

Now, in the ideal world, we'd declare the Go SDK as a dev dependency, and everything would be fine. However, we can't do this because there's no way to gate the call to `register_toolchain` on us being the root module or not.

The workaround? Attempt to use a unique name for the Go SDK that we need and hope that this works as we think it does.

### DIFF
--- a/gazelle/MODULE.bazel
+++ b/gazelle/MODULE.bazel
@@ -9,17 +9,6 @@ bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(name = "gazelle", repo_name = "bazel_gazelle", version = "0.28.0")
 bazel_dep(name = "rules_go", repo_name = "io_bazel_rules_go", version = "0.35.0")
 
-go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
-
-go_sdk.download(
-    name = "go_sdk",
-    version = "1.18",
-)
-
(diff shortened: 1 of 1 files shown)


fmeum:
@shs96c Do you have to add an SDK explicitly? The rules_go module should register an SDK for you that is picked up if no module closer to the root registers a custom one.

shs96c:
Test failures on `last_green` appear to be something to do with test files not being present in the `data` section of `sh_test`. Not sure if it's related, but the test _does_ pass when not using `bzlmod`

fmeum:
@shs96c That's probably on me: https://github.com/bazelbuild/bazel/commit/839ce7f5c40240d8b6f49c416c3769e226f43fee

This was intended to be backwards compatible, I will debug why it isn't.

Wyverald:
@fmeum just to confirm, do you think this change is good? If so, I'll go ahead and merge it

fmeum:
@Wyverald Yup, the change it's good, it just no longer matches the PR description.

Wyverald (approved):

PR #412, opened 2022-11-15
Negation support

leovitch:
This PR adds the ability to negate individual elements of a config_settings_group(), so that you can construct groups that evaluate conditions like "match config_setting 1 and do not match config_setting 2." This is very helpful when making sure that the list of conditions in a select match once and only once.

### DIFF
--- a/docs/selects_doc.md
+++ b/docs/selects_doc.md
@@ -79,7 +79,7 @@ A dictionary usable by a native `select()`.
 ## selects.config_setting_group
 
 <pre>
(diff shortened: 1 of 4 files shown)


leovitch:
FWIW the failure with bzlmod looks similar to the failure discussed on https://github.com/bazelbuild/bazel-skylib/pull/411 where [fmeum](https://github.com/fmeum) wrote: 
@shs96c That's probably on me: https://github.com/bazelbuild/bazel/commit/839ce7f5c40240d8b6f49c416c3769e226f43fee

fmeum:
@leovitch These failures should be fixed by https://github.com/bazelbuild/bazel/pull/16755. Once it has been merged, you can do a trivial amend of your commits and force push to trigger a rerun of the pipeline.

leovitch:
Thanks for the note Fabian!

PR #413, opened 2022-11-15
Add flatten function that collapses first dimension of an iterable into a list.

SamiKalliomaki:


aiuto:
This seems like a very low value thing to add.  It is basically a wrapper for a one line list comprehsension.

SamiKalliomaki:
The use case I have is:
```bazel
flatten([
    select({
        ":%s_enabled" % feature: ["//some/path/%s:deps" % feature],
        "//conditions:default": [],
    })
    for feature in ALL_FEATURES
])
```

IMO that is a lot clearer than:
```bazel
[target for target in sublist for sublist in [
    select({
        ":%s_enabled" % feature: ["//some/path/%s:deps" % feature],
        "//conditions:default": [],
    })
    for feature in ALL_FEATURES
]]
```

aiuto:
What are you actually trying to accomplish with that?

SamiKalliomaki:
I want to allow passing in a command line argument to allow selecting which features to include. Because the number of "features" is large (around 100 in my case) and changes often, it is not feasible to do this without for loops.

SamiKalliomaki:
Actually the list comprehension implementation does not work with selects. Updated implementation.

aiuto:
It's not my call, but I don't think we should be adding more functions to skylib unless they fill a clear need for many users. I won't be approving or merging this. Anyone else?

PR #415, opened 2022-11-25
Add mutable union and difference operations on sets

evertedsphere:
This will enable https://github.com/tweag/rules_haskell/ to remove its custom `set` library altogether, since all the other operations are already implemented in `sets.bzl`.


evertedsphere:
I can't tell what the issue is with the docstrings: the formatting looks identical to the docstrings for the other functions in the `new_sets.bzl` module.

PR #421, opened 2023-01-03
Add extension to export module version as Starlark constant

fmeum:
Fixes https://github.com/bazelbuild/bazel/issues/17114


fmeum:
@Wyverald Could you review this PR (at least for the Bzlmod content)?

extensions/module_version.bzl
```diff
@@ -0,0 +1,63 @@
...
+
+_module_version_repo = repository_rule(
+    implementation = _module_version_repo_impl,
+    attrs = {
+        "version": attr.string(),
+    },
+)
+
+def _module_version_impl(module_ctx):
+    # buildifier: disable=no-effect
```
Wyverald, line 52: why not just use a regular for loop instead?

Wyverald: actually, you also need to dedupe by name here

fmeum: No particular reason, I just thought that list comprehensions were the "idiomatic" way to instantiate rules dynamically. Will change to a for loop.

Could you explain why I need to deduplicate? Is that only because of `multiple_version_override`?

Wyverald: yeah. it's fine to select one version to export when a `multiple_version_override` exists, but you don't want to outright fail the build in such a case

fmeum: Done, I'm picking the one closest to the root.

Wyverald (approved):
looks nice!

fmeum:
CI is failing since the top-level MODULE.bazel file now declares overrides. I can work around this in the CI configuration when the PR is generally accepted.

stonier:
Swinging by to support this PR.

**Use Case**

Shipping a python wheel from a module with the same version as the module itself.

**Current Practice**

The module version string is configured and maintained in two places - 1) `MODULE.bazel` and 2) the `BUILD.bazel` which configures the `py_wheel` target.

**Desired Practice**

Consolidation of the module version number in one place, `MODULE.bazel`.

NB: Implementation looks like it will do the job!

fmeum:
Friendly ping, rules_go and gazelle would also like to use this to learn of each others version.

fmeum:
I will close this now that we have `module_name()` and `module_version()`.

PR #424, opened 2023-01-17
Add gazelle plugin to CI and distribution mechanism

tetromino:
After #400, the gazelle plugin has been cleanly separated out into its own bazel workspace, which will soon finally allow us to mark it stable. But this means:

* we need to change our bazelci config to explicitly build and test it, since `bazel build //...` no longer includes the plugin;
* we need to add proper distribution rules for it;
* we need to update release instructions, since now we will have two distribution tarballs

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,47 +1,44 @@
 ---
+.reusable_config: &reusable_config
+  build_targets:
+    - "--"
+    - "//..."
+    - "@bazel_skylib_gazelle_plugin//..."
+  test_targets:
+    - "--"
+    - "//..."
+    - "@bazel_skylib_gazelle_plugin//..."
+
 tasks:
   ubuntu1804_latest:
+    <<: *reusable_config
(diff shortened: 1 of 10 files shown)


tetromino:
Note: as a followup, I will try to see if we can clean up distro BUILD files to not include distro-only targets. I think this is feasible by updating to a newer version of rules_pkg, but the change is big enough that it would be best done as a separate followup PR.

achew22:
This LGTM. Obviously gotta fixup CI but conceptually this seems like the right direction. Thanks @tetromino!

brandjon (approved):

alexeagle:
This module also needs to be added to the Bazel Central Registry in order to use it - @tetromino is that on your list? Want a PR for it?

tetromino:
@alexeagle - yes, it's my next todo; I just need to generate the tarball

alexeagle:
I guess it's a wider discussion, but I think we could chat about distribution of the entire repo instead, it's tremendously simpler.

PR #430, opened 2023-02-02
Update maintainer guide for bzlmod and bazel-central-registry

tetromino:


brandjon (approved):
That's an intimidating series of steps at a glance, but I guess it can't be helped.

No need to do it in this PR, but it might be nice to separate the release notes template into a separate file. I had trouble initially visually parsing the template vs the surrounding instructions.

tetromino:
As a followup, instead of a long set of instructions, we ought to have (1) a script, and (2) a hook that automatically runs said script after a release is tagged on github. But that's for future maintainers :)

fmeum:
You could use https://github.com/bazel-contrib/publish-to-bcr to automate BCR releases. Many rulesets are already using it.

tetromino:
@fmeum - nice, thank you for the suggestion!

tetromino:
... alas, it seems that publish-to-bcr assumes a 1-1 mapping between git repo and module. Won't work for us since gazelle plugin is a separate module.

tetromino:
https://github.com/bazel-contrib/publish-to-bcr/issues/46

PR #432, opened 2023-02-08
fix: allow empty glob for `testdata` in the Gazelle extension

### LINKED ISSUE #431: Error using Gazelle extension with `--incompatible_disallow_empty_glob` enabled
Error Message
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_chuck/91330c3c30ebaed046a35357439e8c08/external/bazel_skylib_gazelle_plugin/bzl/BUILD", line 28, column 13, in <toplevel>
		] + glob([
Error in glob: glob pattern 'testdata/**' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
ERROR: /Users/chuck/code/cgrindel/bazel-starlib/renovate/bazel_skylib-1.x/BUILD.bazel:15:15: Target '@bazel_skylib_gazelle_plugin//bzl:bzl' contains an error and its package is in error and referenced by '//:gazelle_bin'
```

It looks like the Gazelle extension is packaged without its test data. [This glob](https://github.com/bazelbuild/bazel-skylib/blob/main/gazelle/bzl/BUILD#L28-L30) causes the error when `--incompatible_disallow_empty_glob` is enabled. I suspect that we just need to explicitly set `allow_empty = True` in that glob.

cgrindel:
Closes #431

### DIFF
--- a/gazelle/bzl/BUILD
+++ b/gazelle/bzl/BUILD
@@ -1,5 +1,5 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 load("@bazel_gazelle//:def.bzl", "gazelle", "gazelle_binary")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 # gazelle:exclude testdata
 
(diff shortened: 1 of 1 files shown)


tetromino (approved):
Thanks for the quick fix!

PR #440, opened 2023-03-24
Expose int- and string-valued build settings as Make variables

fmeum:
While build settings allow for much cleaner flag and setting definitions than `--define`, they have the major drawback that rules need to provide dedicated support for them, which isn't the case for native and most community-maintained rules.

This change attempts to bridge this gap by optionally exposing the value of the common build setting types as Make variables to rules that depend on them via the `toolchains` attribute: If the new `make_variable` attribute is set, the value of the flag or setting is available as a Make variable with that.

Consistency with pre-defined Make variables is enforced by limiting the character set for `make_variable` values to `[A-Z0-9_]`. The new attribute is also only added to int- and string-valued build settings as the other types lack a canonical stringification.

Work towards https://github.com/bazelbuild/bazel/issues/14859

### DIFF
--- a/docs/common_settings_doc.md
+++ b/docs/common_settings_doc.md
@@ -50,7 +50,7 @@ A bool-typed build setting that cannot be set on the command line
 ## int_flag
 
 <pre>
-int_flag(<a href="#int_flag-name">name</a>)
+int_flag(<a href="#int_flag-name">name</a>, <a href="#int_flag-make_variable">make_variable</a>)
 </pre>
 
 An int-typed build setting that can be set on the command line
@@ -61,6 +61,7 @@ An int-typed build setting that can be set on the command line
 | Name  | Description | Type | Mandatory | Default |
 | :------------- | :------------- | :------------- | :------------- | :------------- |
 | <a id="int_flag-name"></a>name |  A unique name for this target.   | <a href="https://bazel.build/concepts/labels#target-names">Name</a> | required |  |
+| <a id="int_flag-make_variable"></a>make_variable |  If set, the build setting's value will be available as a Make variable with this name in the attributes of rules that list this build setting in their 'toolchains' attribute.   | String | optional | <code>""</code> |
 
 
 <a id="int_setting"></a>
@@ -68,7 +69,7 @@ An int-typed build setting that can be set on the command line
 ## int_setting
 
 <pre>
-int_setting(<a href="#int_setting-name">name</a>)
+int_setting(<a href="#int_setting-name">name</a>, <a href="#int_setting-make_variable">make_variable</a>)
 </pre>
 
 An int-typed build setting that cannot be set on the command line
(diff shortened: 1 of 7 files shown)


fmeum:
CC @aranguyen for the configurability point of view

fmeum:
@aranguyen @brandjon Friendly ping, this has come up in Slack again. It does solve a real need.

docs/common_settings_doc.md
```diff
@@ -61,14 +61,15 @@ An int-typed build setting that can be set on the command line
...
+| <a id="int_flag-make_variable"></a>make_variable |  If set, the build setting's value will be available as a Make variable with this name in the attributes of rules that list this build setting in their 'toolchains' attribute.   | String | optional | <code>""</code> |
 
 
 <a id="int_setting"></a>
 
 ## int_setting
 
 <pre>
-int_setting(<a href="#int_setting-name">name</a>)
+int_setting(<a href="#int_setting-name">name</a>, <a href="#int_setting-make_variable">make_variable</a>)
```
gregestren, line 72: What's the use case for supporting `--*_setting`, if it can't be set by the user at the command line?

gregestren (commented):
Interesting proposal.

This only works for attributes that support make variable expansion, right? So some but not all native attributes. And only Starlark rules that know how factor it in via `ctx.var`?

The original request is I think asking for a generic mechanism that works everywhere , without rule authors even having to opt in. I realize you're not consciously trying to go that far here, but how do you see this fitting with the broad idea? Do you think the broad idea makes sense? 

I'm fairly sympathetic, and offhand think if we wanted to support the broader idea doing so with Starlark flags probably makes sense?  But a complete generic variation would have to involve some actual Bazel refactoring?

fmeum:
My aim with this PR is to realize the following:

* Make build settings drop-in, fully featured replacements for `--define`s. This requires some kind of interior with existing rules that doesn't require them to be changed and make variables are a natural fit.
* Provide rules that currently support neither `--define`s nor build settings to easily add support that applies even to older versions of Bazel (just use `ctx.expand_make_variables` or any of the Starlark replacements for it).

As a result, users could start replacing `--define`s with build settings, knowing that they won't incur new limitations. If we see increased adoption (even Bazel's IntelliJ plugin still relies on defines), we would have an easier time coming up with more informed designs of new features for build settings.

I don't have a concrete idea for universal support for build settings that doesn't involve explicit rule opt-in. At this point I'm not even sure whether it would be necessary - moving the Starlark versions of Make variable and location expansion used by the built-in rules to Skylib and packaging them in a very friendly API may actually be good enough.

docs/common_settings_doc.md
fmeum: They can still be set via transitions, which combined with this PR makes it possible to control rule attributes with make variable support via transitions.

gregestren:
That all works for me. Thanks.

gregestren (approved):

gregestren:
We'll have to deal with write access reviewers (I'm not. :( ) after my vacation if no one else gets to this first.

Wyverald (approved):

PR #447, opened 2023-05-15
Build with incompatible_disallow_empty_glob

limdor:
In order to flip the flag, all downstream projects should be adapted. However, it is hard to fix them all if there are constant regressions. Adding it to the CI will ensure that once the project can build with incompatible_disallow_empty_glob it can keep building like that. See: bazelbuild/bazel#15327


tetromino (approved):

tetromino:
Thanks, this will help prevent issues like https://github.com/bazelbuild/bazel-skylib/issues/431

PR #448, opened 2023-05-24
Avoid building build_test deps unnecessarily

UebelAndre:
I ran into an issue where I had a `build_test` that was only compatible with a particular platform. I had annotated the target with `target_compatible_with` but continued to get builds on the incompatible platform. This came down to my `bazel test //...` invocation picking up the `{name}_{idx}__deps` targets and building the dependency anyway. This change updates these targets to account for newer common attributes and tags them as manual so they're only built when the user facing test target is built.

### DIFF
--- a/rules/build_test.bzl
+++ b/rules/build_test.bzl
@@ -41,6 +41,14 @@ _empty_test = rule(
     test = True,
 )
 
+_GENRULE_ATTRS = [
+    "compatible_with", 
+    "exec_compatible_with",
+    "restricted_to", 
+    "tags", 
+    "target_compatible_with",
+]
+
 def build_test(name, targets, **kwargs):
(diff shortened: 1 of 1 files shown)


tetromino (approved):
Looks good, thanks. I pushed a whitespace fix to make the buildifier check pass.

UebelAndre:
Thank you!

UebelAndre:
@tetromino would it be possible to also get a new release after this change?

UebelAndre:
@tetromino friendly ping 😅

tetromino:
> @tetromino friendly ping sweat_smile

Draft PR: https://github.com/bazelbuild/bazel-skylib/pull/451

(Note we also may need to fix some sort of stardoc-related weirdness with bazel at head to unblock ci.)

PR #450, opened 2023-05-31
Allow .scl files in bzl_library

tetromino:
See bazelbuild/bazel@a0cd355

### DIFF
--- a/bzl_library.bzl
+++ b/bzl_library.bzl
@@ -46,11 +46,11 @@ bzl_library = rule(
     implementation = _bzl_library_impl,
     attrs = {
         "srcs": attr.label_list(
-            allow_files = [".bzl"],
-            doc = "List of `.bzl` files that are processed to create this target.",
+            allow_files = [".bzl", ".scl"],
+            doc = "List of `.bzl` or `.scl` files that are processed to create this target.",
         ),
         "deps": attr.label_list(
-            allow_files = [".bzl"],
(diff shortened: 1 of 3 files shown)


bzl_library.bzl
```diff
@@ -46,19 +46,19 @@ bzl_library = rule(
     implementation = _bzl_library_impl,
     attrs = {
         "srcs": attr.label_list(
-            allow_files = [".bzl"],
-            doc = "List of `.bzl` files that are processed to create this target.",
+            allow_files = [".bzl", ".scl"],
+            doc = "List of `.bzl` or `.scl` files that are processed to create this target.",
```
brandjon, line 50: Nit: Here and in rule doc: Maybe "`.bzl` and `.scl`" -- "and" instead of "or" to make clear it's not necessarily a homogeneous list of only .bzl or only .scl.

tetromino: Good point.

tetromino:
CI failure with Bazel HEAD (https://buildkite.com/bazel/bazel-skylib/builds/2694) is due to a java toolchain change in https://github.com/bazelbuild/bazel/commit/975866a3e09212766bd353a90a8c42efe91d53ab which broke Stardoc.

To fix, we need to update to the latest Stardoc release (0.5.6) - for which we need https://github.com/bazelbuild/bazel-central-registry/pull/677 to be merged first.

PR #455, opened 2023-08-28
Modify actions in order not to need `toolchain` param 

kotlaja:
All actions which use tool or executable for which is not clear if it comes from a toolchain, must set a `toolchain` parameter ( migration of Automatic Exec Groups). 

As we discussed internally, I've modified actions so that it's recognised that tools are not from the toolchain. Hence, there will not be an error which states `Couldn't identify if tools are from implicit dependencies or a toolchain. Please set the toolchain parameter. If you're not using a toolchain, set it to 'None'.`. Hence, no need for the toolchain parameter.

### DIFF
--- a/distribution/BUILD
+++ b/distribution/BUILD
@@ -1,7 +1,7 @@
-load("distribution.bzl", "remove_internal_only")
 load("@bazel_skylib//:version.bzl", "version")
-load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@rules_pkg//:mappings.bzl", "pkg_files", "strip_prefix")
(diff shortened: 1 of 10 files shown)


comius (approved):

tetromino (commented):
Question: does this break compatibility with older Bazel releases (before 6.3.2)?

If it does, that is not necessarily a blocker, but it is something that we must note in changelog and release notes, and it would affect what version number the next release should have.

tetromino (approved):
I've manually verified tests still pass on Bazel 5.4.1.

Took the liberty of pushing a commit to fix the buildifier lint error.

PR #457, opened 2023-09-05
Add error for empty bzl_library

keith:
This improves the error in the case your bzl_library does nothing. Otherwise you end up with something weirder later:

```
BUILD:35:12: in deps attribute of bzl_library rule LABEL: 'DEP' does not produce any bzl_library deps files (expected .bzl)
```

Ideally we could set `allow_empty = False` on `srcs` but currently it's valid to just have a bzl_library target that aggregates multiple other libraries in its deps.


comius (approved):

keith:
@comius can you land this?

comius:
> @comius can you land this?

Build failures are happening at head and are not related to this change.

Before releasing skyline we should fix those.

PR #463, opened 2023-09-20
versions: Don't fail on Bazel dev builds

fmeum:
Dev builds of Bazel are assumed to be more recent than any released version.

### DIFF
--- a/docs/versions_doc.md
+++ b/docs/versions_doc.md
@@ -26,6 +26,9 @@ Parses a version string into a 3-tuple of ints
 
 int tuples can be compared directly using binary operators (&lt;, &gt;).
 
+For a development build of Bazel, this returns an unspecified version tuple
+that compares higher than any released version.
+
 
 **PARAMETERS**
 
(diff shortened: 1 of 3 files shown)


fmeum:
@Wyverald Could you review? It's almost identical to https://github.com/bazel-contrib/bazel_features/blob/541c6ac87a93c95df76388a6f1d0b3f65d02fbac/private/parse.bzl#L19C1-L20C1.

Wyverald (approved):

fmeum:
It would be great if we could have a new Skylib release, which would also make https://github.com/bazelbuild/bazel-skylib/commit/8386b9d32bf69dd2d2f92d9ca39582cf6dabeb37 and https://github.com/bazelbuild/bazel-skylib/commit/0171c69e5cc691e2d0cd9f3f3e4c3bf112370ca2 available.

fmeum:
@brandjon Would it be possible to release a new version of Skylib with these fixes?

PR #464, opened 2023-09-20
Add `doc` param to `unittest.make`

UebelAndre:
Same change as https://github.com/bazelbuild/bazel-skylib/pull/343 but now for `unittest`


UebelAndre:
@comius friendly ping here 😄

comius (approved):

comius:
The breakages are not related to this change.

PR #467, opened 2023-10-09
Add option to include runfiles files in the `select_file` search

### LINKED ISSUE #466: Search runfiles in `select_file` rule
`select_file` rule currently only searches within `DefaultInfo.files`.

I was wondering if we can add an argument to `select_file` to optionally expand the search criteria to include `DefaultInfo.default_runfiles`

It would look something like this.
```starlark
select_file(
    name = "selected",
    srcs = ":files",
    subpath = "foo/bar.py",
    include_runfiles = True,
)
```

kpark-hrp:
Adds `include_runfiles` boolean argument to the `select_file` rule.

If set to `True`, runfiles files of the target is also searched for the matching `subpath` file.

resolves #466

### DIFF
--- a/docs/select_file_doc.md
+++ b/docs/select_file_doc.md
@@ -11,7 +11,7 @@ Selects a single file from the outputs of a target by given relative path.
 ## select_file
 
 <pre>
-select_file(<a href="#select_file-name">name</a>, <a href="#select_file-srcs">srcs</a>, <a href="#select_file-subpath">subpath</a>)
+select_file(<a href="#select_file-name">name</a>, <a href="#select_file-include_runfiles">include_runfiles</a>, <a href="#select_file-srcs">srcs</a>, <a href="#select_file-subpath">subpath</a>)
 </pre>
 
 Selects a single file from the outputs of a target by given relative path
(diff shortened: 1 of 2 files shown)


kpark-hrp:
Hello team, @Wyverald @brandjon @tetromino 

Can I get a review?

Wyverald:
@tetromino is on leave until next year. I'm rather unfamiliar with this part of the code, so I'll leave the review to @brandjon.

kpark-hrp:
I see. @brandjon Can you take a look at this small PR?

The default behavior doesn't change. It just adds an option.

kpark-hrp:
@Wyverald Hey, is it possible for you to review or let @brandjon know personally?

comius:
Skylib is one of the core libraries at least I think so. I believe it should have a higher bar to introduce new features. If there were more users requesting such a feature, I'd be more open to introducing it.

Also, PR lacks testing and support for data runfiles or symlinks that runfiles can have.

kpark-hrp:
1. Bazel doc says avoid using `data_runfiles`. https://bazel.build/rules/lib/providers/DefaultInfo#data_runfiles
2. Can symlink support be added later if there is a request for it. I am currently only searching from the `files` parameter of the `runfiles` https://bazel.build/rules/lib/builtins/runfiles.html

@comius I can add a test or two for along with this PR. It would be nice to have this merged to the upstream, so I don't have to maintain a fork or patch of this skylib rule. Also, this doesn't change the default behavior of this rule, so I think it's rather harmless.

comius:
I have a strong argument against this change. This flattens a depset, and we should avoid that if possible. Specially in rules coming with skylib.

Issue #399, opened 2022-10-06
Feature request: Make "out" parameter of "native_binary" optional


carlosgalvezp:
Hi,

We find ourselves duplicating a lot of code for no benefit when using `native_binary`:

```
native_binary(
  name = "foo",
  src = "path/to/foo",
  out = "path/to/foo",
)
```

Would it be possible to make the `out` parameter optional, and if not set it should be set to the same path as the `src`? 

Thanks!

tpudlik:
`src` is actually not suitable as a default because it may be in a different package than the `native_binary` target (and that's not allowed for a rule output). But I think setting the default to `name` would work. I'll give that a try.

PR #474, opened 2023-11-06
Make `out` on `native_binary` optional

tpudlik:
Fixes: #399


tpudlik:
We could avoid the `.exe` suffix on non-Windows machines by adding a `is_windows` attribute populated from a `select` in a wrapper macro (like in https://github.com/bazelbuild/bazel-skylib/blob/main/rules/private/copy_file_private.bzl#L140-L143), but adding a wrapper macro for such a minor aesthetic benefit seems a little overkill. Let me know if you'd like me to do it, though!

tetromino (approved):
Thanks!

PR #476, opened 2023-11-14
Remove outdated rules_cc dependency

### LINKED ISSUE #475: Brings in very old version of rules_cc: Eliminate or update?
Hi wonderful Bazelers,

I was debugging something else and noticed that skylib brings in a very old version of rules_cc, breaking some things.

My understanding is that the dependency could probably be eliminated, since the cc rules used are built-in anyway--but failing that perhaps it would be worth an update?

Lmk what you think. Happy to help (though it might just be fastest to make the change, once you decide which makes more sense).

Thanks so much for all you do!
Chris
(ex-Googler)

cpsauer:
Fixes #475 
^ please see that issue for context before reading :)


keith (approved):

cpsauer:
(Sorry for the accidental/unnecessary close and reopen, but to clarify, this is ready for a review&merge.)

cpsauer:
Friendly bump on review/merge since it's been 1mo :)

tetromino (approved):
Thanks!

PR #478, opened 2023-11-15
Modify actions in order not to need toolchain param

kotlaja:
The same as https://github.com/bazelbuild/bazel-skylib/pull/455.


c-mita (approved):
LGTM

comius (approved):

PR #484, opened 2024-01-11
Add `env` and `env_inherit` to `native_binary` and `native_test`

redsun82:
This relies on the `RunEnvironmentInfo` that was introduced in bazel 5.3.0. To keep the library compatible with previous versions,  [`bazel_features`](https://github.com/bazel-contrib/bazel_features) is used.

To make it possible to run CI and update the docs without all the dependencies from `@bazel_features` (some of which would be private and not visible to the doc generation rule), a `bazel_features_stub` directory is introduced which remains unused when using a released version of the library (it gets removed by `MODULE.bazel-remove-override.patch`).

Also, this requires a user to do an additional `WORKSPACE` setup step if they need to use `native_binary.bzl` and they are not using bzlmod.


jgertm (approved):

hvadehra:
> This relies on the RunEnvironmentInfo that was introduced in bazel 5.3.0. To keep the library compatible with previous versions, [bazel_features](https://github.com/bazel-contrib/bazel_features) is used.

FYI, going by https://bazel.build/release#support-matrix , v5.4.1 is now the minimum supported version, so if `RunEnvironmentInfo` is the only thing needed for this, I'd say it's safe to just assume it's available, and you can drop the dependency on `bazel_features` & related complexity.

(not a full review, I'll defer to @tetromino / @brandjon for that)

redsun82:
@hvadehra thanks for the heads up, that makes things _way_ easier. @tetromino / @brandjon, the PR should be ready for another review pass now :rocket:

tests/native_binary/assertenv.cc
```diff
@@ -0,0 +1,44 @@
...
+#include <string.h>
+
+#ifdef _WIN32
+#define OS_VAR "APPDATA"
+#define strncasecmp _strnicmp
+#else
+#define OS_VAR "HOME"
+#endif
+
+enum vars_to_be_found {
```
comius, line 11: Could you implement this test in a less hacky way?

For example, get rid of the enum, and just search for 3 strings. Use set found to true/false.

comius (changes requested):
LGTM (except for the test)

redsun82:
@comius sorry for the late update on this, I've cleaned up the test now.

PR #486, opened 2024-01-23
Adding `expansion.bzl` for Fully Expanding Environment Variables

CauhxMilloy:
The built-in functionality (exposed Skylark methods on `ctx`) for expanding environment variables leaves a lot of implementation to do in `bzl` files. This PR adds in that missing functionality for easy reuse with a new `expansion` struct. `expansion` has various methods for expanding environment variables with a flexible API.

The existing APIs handle only one piece at a time:
* `ctx.expand_location()` only handles `$(location ...)` (and similar) expansion. It does not handle any "make variable" expansion (no expansion from `TemplateVariableInfo` or other providers via toolchains).
* `ctx.expand_make_variables()` only handles make variables. If it encounters `$(location ...)` (or similar), it [errors out with no means of recovery](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/ConfigurationMakeVariableContext.java#L136). This method is also marked as deprecated, with `ctx.var` listed as the preferred API.
* `ctx.var` is a simple dictionary, which contains resolved mappings based on toolchains. However, being a simple data structure (not a function) leaves recursive functionality and/or integration with `ctx.expand_location()` to be implemented by hand (or in this PR).

Many internal systems make use of functionality that fully resolves both make variables and `$(location ...)` (and does so recursively). This is done with `ruleContext.getExpander().withDataLocations()` ([example](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java#L522)). However, this is never exposed to skylark. This means that built-in rules will have the `env` attribute fully expanded, but custom rules cannot (easily).

The above mentioned methods have their own (somewhat duplicated) implementations ([`ctx.expand_location()`](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java#L1011), [`ctx.expand_make_variables()`](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java#L964), [`ctx.var`](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java#L823)). Note the identical `ConfigurationMakeVariableContext` initialization [here](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java#L949) and [here](https://github.com/bazelbuild/bazel/blob/36fa60b1805faa7da2c4b5330b4b186740f5f00d/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java#L978) -- identical except for the use of `additionalSubstitutionsMap`, which could have been delegated (similar to the impl of `LocationTemplateContext`). Also note the separate/duplicated parsing implementations in [`LocationTemplateContext`](https://github.com/bazelbuild/bazel/blob/addf41bce1f26e8bc4bd0b09cb2fbffbb0446f25/src/main/java/com/google/devtools/build/lib/analysis/stringtemplate/TemplateExpander.java#L77) and in [`LocationExpander`](https://github.com/bazelbuild/bazel/blob/36fa60b1805faa7da2c4b5330b4b186740f5f00d/src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java#L179).

This PR tries to avoid more duplicate implementations (which can't really happen anyway; as this is in external Skylark, not Java / Bazel runtime). These new methods make use of `ctx.expand_location()` and `ctx.var` to allow fully recursive variable expansion that incorporates both inputs (toolchains and `location`). The methods avoid copies/string mutations/extra loops when necessary to ensure that the functions can run quickly. The methods' API allows for manual/direct data structures (lists/dicts) as input, or for pulling values directly from `ctx`. `tests/expansion_tests.bzl` added to test all added methods.

This PR is being submitted here so that it can be reused (instead of copied in many repos). It is also preferable to add functionality here in Skylib, instead of (or in addition to) any changes in the [Bazel repo](https://github.com/bazelbuild/bazel) so that it can be more easily pulled into projects with a pinned Bazel version.

Please feel free to leave comments or suggestion on ways to improve this PR, or let me know if you have any questions. Thanks!

### DIFF
--- a/lib/expansion.bzl
+++ b/lib/expansion.bzl
@@ -1,0 +1,338 @@
+
+"""Skylib module containing functions that aid in environment variable expansion."""
+
+def _valid_char_for_env_var_name(char):
+    return char.isalnum() or char == "_"
+
+def _key_to_be_expanded(str_with_key, key, start_of_key_index):
+    # Check that the string at index is prefixed with an odd number of `$`.
+    # Odd number means that the last `$` is not escaped.
+    dollar_sign_count = 0
+    for index in range(start_of_key_index, -1, -1):
+        if str_with_key[index] != "$":
+          break
+        dollar_sign_count += 1
+
+    # Check that the key is correctly matched.
+    # Specifically, check the key isn't matching to another key (substring).
+    key_mismatch = False
+    if key[-1] not in (")", "}"):
+        end_of_key_index = start_of_key_index + len(key)
+        key_mismatch = (
+            (end_of_key_index < len(str_with_key)) and
+            _valid_char_for_env_var_name(str_with_key[end_of_key_index])
+        )
+
+    return (not key_mismatch) and (dollar_sign_count % 2) == 1
+
+def _expand_key_in_str(key, val, unexpanded_str):
+    key_len = len(key)
+    val_len = len(val)
+    searched_index = 0
+    expanded_str = unexpanded_str
+    # Max iterations at the length of the str; will likely break out earlier.
+    for _ in range(len(expanded_str)):
+        used_key_index = expanded_str.find(key, searched_index)
+        if used_key_index < 0:
+            break
+        if _key_to_be_expanded(expanded_str, key, used_key_index):
+            # Only replace this one instance that we have verified (count = 1).
+            # Avoid extra string splicing, if possible.
+            if searched_index == 0:
+                expanded_str = expanded_str.replace(key, val, 1)
+            else:
+                expanded_str = (
+                    expanded_str[0:searched_index-1] +
+                    expanded_str[searched_index:].replace(key, val, 1)
+                )
+            searched_index += val_len
+        else:
+            searched_index += key_len
+    return expanded_str
+
+def _expand_all_keys_in_str_from_dict(replacement_dict, unexpanded_str):
+    # Manually expand variables based on the var dict.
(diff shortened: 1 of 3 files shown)


comius:
Wow, this is one nice PR. It's a shame it's here for 1.5 years.

peakschris:
I need this. Can it be merged? Or maybe moved to https://github.com/bazel-contrib/bazel-lib?

CauhxMilloy:
> I need this. Can it be merged? Or maybe moved to https://github.com/bazel-contrib/bazel-lib?

Funny that you mention this. I was planning on moving this PR to bazel-contrib/bazel-lib in the hopes that it'd go somewhere. I'll put that up once I find the time. (Also, I have a few more edits to make.)

PR #491, opened 2024-03-06
Support custom `mnemonic` and `progress_message`

joshua-k-harris:
Adds support for adding a custom `mnemonic` and `progress_message` for `copy_file`, `copy_directory` and `run_binary` rules.


tetromino (changes requested):
Looks reasonable overall.

Out of curiosity, why in the macros is the order of args `progress_message, mnemonic` instead of the other way around (alphabetically)?

It would probably be good to add test for this. There isn't any conveniet way to test the progress message, but we can test the mnemonic via skylib's own unittest.bzl (retrieve via `analysistest.target_actions(ctx)[0].mnemonic`)

comius:
cc @brandjon 

Setting mnemonics from targets is something we had huge problems with internally.

joshua-k-harris:
> cc @brandjon
> 
> Setting mnemonics from targets is something we had huge problems with internally.

Does this impact the inclusion of this change or can you restrict the use of the attributes internally?

comius:
>> Setting mnemonics from targets is something we had huge problems with internally.

> Does this impact the inclusion of this change or can you restrict the use of the attributes internally?

We can't restrict them internally - we only monitor them :(.

The attributes should be restricted in Bazel and this would be a very welcome contribution. The idea is, that ctx.actions.run(mnemonic) can only be set to a literal string. This way you can still use your own mnemonics, but you're more limited and you can't do `for i in range(10000000): my_macro(mnemonic=i)`. (Creating 1M different mnemonics happened to us). Technically Starlark already has a bag of all the literals, because it's interning them. One would need to check that mnemonics is in that bag.

This PR is acceptable after the PR in Bazel.

sitaktif:
Thanks @comius for the context. I opened https://github.com/bazelbuild/bazel/issues/23575 on the Bazel side.

PR #492, opened 2024-03-14
Support passing arguments to run_binary via param file

vejbomar:
Allows to use `run_binary` in the case when the size of the command line can grow longer than the maximum size allowed by the system. We're currently hitting command line limits in our build setup.

Implementation inspired from [Scala rules](https://github.com/bazelbuild/rules_scala/blob/v6.4.0/scala/private/rule_impls.bzl#L77).


rules/run_binary.bzl
```diff
@@ -92,5 +97,13 @@ run_binary = rule(
...
                   " expansion.",
         ),
+        "param_file_format": attr.string(
+            doc = "If provided, Bazel can pass command-line arguments to the tool via file.\n\n" +
+                  "Should be used when the size of the command line can grow longer than the " +
+                  "maximum size allowed by the system. The format is the same as the format of " +
+                  "`param_file_arg` in [`Args.param_file_arg`](https://bazel.build/rules/lib/builtins/Args#use_param_file). " +
+                  "The `tool` has to support reading arguments from the file for this to work.",
+            default = "",
+        ),
```
comius, line 107: I had hard time understanding this and I needed to open Bazel's documentation. Although I used the functionality before.

I'd suggest just pasting the Bazel's doc here. Maybe adding that param files are only used if the attribute is set.

```
A format string with a single "%s". If the args are spilled to a params file then they are replaced with an argument consisting of this string formatted with the path of the params file.
For example, if the args are spilled to a params file "params.txt", then specifying "--file=%s" would cause the action command line to contain "--file=params.txt".
```

comius (changes requested):

PR #493, opened 2024-03-20
expand_template: do not output_to_genfiles

tetromino:
output_to_genfiles is deprecated unless needed for backwards compatibility with legacy rules (see https://bazel.build/rules/lib/globals/bzl#rule)


tjgq (approved):
Thanks!

PR #495, opened 2024-04-23
Allow `bzl_library` to depend on non-`bzl_library` targets

Wyverald:
Notably, `filegroup`. `bzl_library` doesn't actually read anything from the `StarlarkLibraryInfo` provider, and requiring all deps to be other `bzl_library` targets is really painful for anyone loading .bzls from `@bazel_tools` or `@platforms` because those core modules/repos don't want a dependency on Skylib just for access to `bzl_library`.

The medium-term plan will be to move `bzl_library` into `@bazel_tools`; but before then, this can serve as a stop-gap.


tetromino (changes requested):
As far as I can see, the test doesn't test the new change. How about this setup instead:

c.bzl loads symbol A from a.bzl and symbol B from b.bzl

a.bzl is in a_filegroup
b.bzl is wrapped in b_bzl_library
c.bzl is wrapped in c_bzl_library with a_filegroup and b_bzl_filegroup in deps

And the test is a rule which verifies that c_bzl_library's DefaultInfo includes a.bzl, b.bzl, and c.bzl

Wyverald:
> As far as I can see, the test doesn't test the new change.

doesn't `:b_fg` (which depends on `:a_fg`, a `filegroup`) test it?

tetromino:
> > As far as I can see, the test doesn't test the new change.
> 
> doesn't `:b_fg` (which depends on `:a_fg`, a `filegroup`) test it?

It doesn't test the content of the providers.

If you want, I can push a commit to this PR to make the test cover this.

Wyverald:
> If you want, I can push a commit to this PR to make the test cover this.

if you have time, that'd be great! otherwise, I can get to it later today.

tetromino:
> Done now.

I was in the process of writing a slightly more universal variation on the same theme, using skylib's standard analysistest machinery; I pushed that since I think it provides a useful example for analysistest usage.

tetromino (approved):

PR #496, opened 2024-04-24
Prepare release 1.6.0

tetromino:
And take the opportunity to reformat docs using a more modern Stardoc release


Wyverald (approved):

fmeum:
The publishing step for the BCR doesn't seem to have run for 1.6.0.

@tetromino You should have received an email with the error.

Wyverald:
I don't think we have publish-to-bcr for bazel-skylib.

PR #456, opened 2023-08-29
Add helper functions for module extensions as `modules`

fmeum:
Adds a new module `modules` with two helper functions for module extensions:

* `use_all_repos` makes it easy to return an appropriate `extension_metadata` from a module extension (if supported) to indicate that all repositories generated by the extension should be imported via `use_repo`.
* `as_extension` turns a WORKSPACE macro into a module extension that uses `use_all_repos` to automate the generation of `use_repo` calls.

PR #497, opened 2024-04-24
Mark `modules.as_extension` as `reproducible`

fmeum:
WORKSPACE macros are fully deterministic and thus a prime example of a `reproducible` module extension. This reduces clutter in `MODULE.bazel.lock`.


fmeum:
@tetromino Could we get this into 1.6.0? It's a follow-up to #456 that, to be honest, I had given up on in the meantime.

CC @Wyverald for a review from the Bzlmod perspective.

Wyverald (approved):

tetromino:
Maybe I'm missing something obvious, but how can you a priori say that a macro is deterministic when it may fetch from a non-deterministic external resource? Consider, for example, an http_archive pointing at some git repo's master HEAD.

PR #499, opened 2024-04-25
Improve compatibility with Bazel 6.

phst:
In Bazel 6, the doc parameter of the module_extension function doesn’t accept None; cf. https://bazel.build/versions/6.5.0/rules/lib/globals#module_extension vs. https://bazel.build/versions/7.0.0/rules/lib/globals/bzl#module_extension. Instead of passing None in that case, just don’t pass the argument at all and use the respective default.


fmeum:
Cc @Wyverald

PR #504, opened 2024-04-25
Add missing `licenses()` and `default_applicable_license` to all BUILD files

tetromino:
Required by internal license checker :/

### DIFF
--- a/gazelle/workspace.bzl
+++ b/gazelle/workspace.bzl
@@ -24,6 +24,16 @@ def bazel_skylib_gazelle_plugin_workspace():
 
     maybe(
         http_archive,
+        name = "rules_license",
+        sha256 = "4531deccb913639c30e5c7512a054d5d875698daeb75d8cf90f284375fe7c360",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/rules_license/releases/download/0.0.7/rules_license-0.0.7.tar.gz",
+            "https://github.com/bazelbuild/rules_license/releases/download/0.0.7/rules_license-0.0.7.tar.gz",
+        ],
+    )
+
+    maybe(
+        http_archive,
         name = "io_bazel_rules_go",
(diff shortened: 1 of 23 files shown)


docs/private/BUILD
```diff
@@ -1 +1,5 @@
-# No targets in this package
+# This package only contains source targets
+
+package(default_applicable_licenses = ["//:license"])
+
+licenses(["notice"])
```
aiuto, line 5: We don't need licenses notice.  Only google uses it and it doesn't mean anything there any more.

gazelle/workspace.bzl
```diff
@@ -22,6 +22,16 @@ def bazel_skylib_gazelle_plugin_workspace():
     """Loads dependencies required to use skylib's gazelle plugin"""
     bazel_skylib_workspace()
 
+    maybe(
+        http_archive,
+        name = "rules_license",
+        sha256 = "4531deccb913639c30e5c7512a054d5d875698daeb75d8cf90f284375fe7c360",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/rules_license/releases/download/0.0.7/rules_license-0.0.7.tar.gz",
```
aiuto, line 30: Does this still exist? I'm not sure we are pushing to that any more as we get more bzlmod'y

aiuto (commented):
We should take this opportunity to drop licenses([notice])

gazelle/workspace.bzl
tetromino: At least for 0.0.7, it works

docs/private/BUILD
tetromino: The internal licenserules doc currently says "The transition to the new license() rules is not yet complete, so both kinds of declarations are still necessary." :/

aiuto: No one is updating that doc.  It only matters if it is licensees(restricted)

tetromino: Interesting! In that case, let's remove `licenses()`.

brandjon (approved):

PR #510, opened 2024-05-10
Copy rules_directory to bazel-skylib.

matts1:
Original implementation is at https://github.com/matts1/rules_directory


matts1:
This is a rather large commit since it's just a copy-paste from rules_directory. I can split it up if that's required, but since that requires some effort on my part I'll just send it as one commit for now, and I can split it up if that's asked of me.

rules/directory/providers.bzl
```diff
@@ -0,0 +1,46 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
```
comius, line 1: Providers should be under `./private` directory, to make it clear, it's an implementation detail of directory rules. Probably also utils.bzl

rules/directory/glob.bzl
```diff
@@ -0,0 +1,87 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
```
comius, line 1: I wish to justify the cost of implementation and mainteinance of a second implementation of glob, but this time Starlark.

So this Starlark glob, allows you to define a directory in one package, and then run glob over it in different package. Are there other differences?

Using using a regular filegroup and a regular glob, you could just merge srcs from the directory rule, with the data from glob rules, right?

comius (changes requested):

rules/directory/providers.bzl
matts1: The providers are the API for this feature. Otherwise, we have no way of writing:
```
cc_toolchain = rule(
  attrs = {
    "sysroot": attr.label(providers = [DirectoryInfo]),
  }
)
```

rules/directory/directory.bzl
```diff
@@ -0,0 +1,173 @@
...
+        transitive_files = depset(
+            direct = sorted(dir_metadata.files, key = lambda f: f.basename),
+            transitive = [
+                d.transitive_files
+                for d in directories.values()
+            ],
+            order = "preorder",
+        )
+        directory = DirectoryInfo(
+            entries = struct(**entries),
```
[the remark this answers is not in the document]
matts1: I was instructed to do this as a solution to the problem of it not being allowed in a depset, but I'll try out a solution that has no depsets.

rules/directory/directory.bzl
```diff
@@ -0,0 +1,173 @@
...
+
+    topological = [root_metadata]
+    for src in ctx.files.srcs:
+        prefix = source_prefix if src.is_source else generated_prefix
+        if not src.path.startswith(prefix):
+            fail("{path} is not contained within {prefix}".format(
+                path = src.path,
+                prefix = root_metadata.human_readable,
+            ))
+        relative = src.path[len(prefix):].split("/")
```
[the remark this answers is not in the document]
matts1: The prefix is not always `src.root`. In the tests in the example, prefix is `.../bazel-skylib/tests/directory/testdata`, while the short path is the path relative to the root (`tests/directory/testdata`).

rules/directory/filter.bzl
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Rules to filter files from a directory."""
+
+load(":providers.bzl", "DirectoryInfo")
+load(":utils.bzl", "get_relative")
+
+def _directory_filter_impl(ctx):
+    directory = ctx.attr.directory[DirectoryInfo]
+    srcs = depset([
```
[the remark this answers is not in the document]
matts1: After my refactor away from depsets, the implementation of this is basically just:
`srcs = ["foo/bar"]` -> `directory.entries["foo"].entries["bar"]`

IMO, this is pretty reasonable, since this allows us to provide nice errors such as distinguishing between "bar doesn't exist within foo" and "the directory foo doesn't exist".

rules/directory/glob.bzl
matts1: This is the use case I have for it.

With this mechanism:
```
# My module extension
http_archive(name = "amd64_sysroot", url=".../amd64_sysroot.tar.gz")
http_archive(name = "arm64_sysroot", url=".../arm64_sysroot.tar.gz", build_file="BUILD.sysroot.bazel")
```

```
# BUILD.sysroot.bazel
directory(name = "sysroot", srcs = glob(["**/*"], exclude=["BUILD.bazel"])
```

```
# toolchain/BUILD.bazel

alias(
    name = "sysroot",
    actual = select({
        "@platforms//arch:amd64": "@amd64_sysroot//:sysroot",
        "@platforms//arch:arm64": "@arm64_sysroot//:sysroot"
        # Other sysroots (eg. RISK-V)
    })
)

subdirectory(
    name = "include_dir",
    directory = ":sysroot",
    path = "usr/include",
)

directory_glob(
    name = "shared_libraries",
    directory = ":sysroot",
    srcs = ["lib/*.so", "lib64/*.so"]
)

cc_sysroot_tool(
    name = "clang",
    directory = ":sysroot",
    executable = "usr/bin/clang",
)

cc_action_type_config(
    name = "compile_config",
    tool = ":clang",
    data = [":include_dir"],
)

cc_action_type_config(
    name = "link_config",
    tool = ":linker",
    data = [":shared_libraries"],
)

cc_toolchain(
    sysroot = ":sysroot",
    include_directories = [":include_dir"],
    action_type_configs = [":compile_config", ":link_config"]
)
```

Without it, toolchains would be significantly harder to define and understand, since every time you want to use anything related to toolchains, you would need to `select` over all possible sysroots.

For example, our rust toolchain would have to write:
```
rust_toolchain(
    ...,
    rustc = select({
        "@platforms//arch:amd64": "@amd64_sysroot//:rustc",
        ...
    })
```

This has two primary downsides:
1) If you ever decide you want to support another platform, with my design, I only have to update two lines (add the sysroot http_archive, and a condition to the select). Without it, I would have to update my whole codebase.
2) (this one is more minor) The top-level build file in `@*_sysroot` gets very massive without it, and there's not an easy way to organize it. Almost all this code would need to go into `BUILD.sysroot.bazel`, while in the current design I can split it up. Of course, you can somewhat mitigate this by providing extra build files, but then it won't work with http_archive.

rules/directory/directory.bzl
```diff
@@ -0,0 +1,173 @@
...
+        human_readable = "@@%s//%s" % (
+            # repo_name is only available in bazel 7+
+            ctx.label.repo_name if hasattr(ctx.label, "repo_name") else ctx.label.workspace_name,
+            ctx.label.package,
+        ),
+    )
+
+    topological = [root_metadata]
+    for src in ctx.files.srcs:
+        prefix = source_prefix if src.is_source else generated_prefix
```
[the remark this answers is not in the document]
comius: You could get away with `bin_dir`, we're in the process of removing `genfiles_dir` . 

Still using `src.root.path` would have less assumptions in the code. At the moment you're assuming files come from one of 2 roots. Technically a tree artifact may also be a root, at the moment we don't allow them in attributes.

rules/directory/glob.bzl
comius: You could put the whole content of ` toolchain/BUILD.bazel` into `BUILD.sysroot.bazel` and use `filegroup` rules like:

```
filegroup(
    name = "include_dir",
    srcs = glob(["usr/include/**"]),
)

filegroup(
    name = "shared_libraries",
    srcs = glob(["lib/*.so", "lib64/*.so"]),
)

cc_sysroot_tool(
    name = "clang",
    executable = "usr/bin/clang",
)

cc_action_type_config(
    name = "compile_config",
    tool = ":clang",
    data = [":include_dir"],
)

cc_action_type_config(
    name = "link_config",
    tool = ":linker",
    data = [":shared_libraries"],
)

cc_toolchain(
    include_directories = [":include_dir"],
    action_type_configs = [":compile_config", ":link_config"]
)
```

Then you can do the selection based on the `cc_toolchain` if need be, or better yet use toolchain resolution.

I'm concerned about out of package globs, that used to be possible with `Fileset`s. Your implementation is a lot better though, because you're only projecting/selecting specific files from a large collection. It might be prone to more rebuilding (change one file and all the targets could be affected), but changes in prebuilt binary tools are probably unlikely.

The main benefit that I see is simplicity - you're using standard rules, and there's no maintenance costs associated with a new rule set. You could also address differences in directory structures among different platforms without introducing new selects.

rules/directory/directory.bzl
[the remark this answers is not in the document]
matts1: I investigated the plausibility of it, and I don't think changing it to `src.root.path` works. What I'm about to say I'm not 100% certain on, so please correct me if I'm wrong:

Consider the following example in blaze:
```
# foo/BUILD
directory(..., srcs = ["source_file", ":genfile", ":binfile"])
```

These three files will have the following respective paths:
```
foo/source_file (root is "")
blaze-out/.../foo/binfile (root is "blaze-out/...")
blaze-genfiles/.../foo/genfile (root is "blaze-genfiles/...")
```

There are now three things we need to do:
1) Check that the file is contained within a the package.
1) Calculate the relative path of the file to the package
1) Generate a `DirectoryInfo` provider with appropriate fields

#### Step 1
My existing solution solves this by:
* Passing in "." (the current directory, which is always a source file) to the directory rule to get the source prefix
* Creating a generated file and getting the directory name of that
* Checking if the source path starts with that prefix

This clearly works for source_file and binfile, but doesn't work for genfile. If I understand correctly, your proposal is to instead use `src.root.path`. However, I don't think this will work because:
* `src.root.path` for `genfile` is `blaze-genfiles/...`
* `src.root.path` for a file in an unrelated repository (eg. `@bar//foo:genfile`, when your directory is in `@foo//foo`) in bazel is `external/bar`

Thus, attempting to strip `src.root.path` will ensure that both these examples result in `foo/genfile`. However, one is valid and the other is invalid.

#### Step 2
This part is easy if you can solve step 1, I don't really care about this.

#### Step 3
Here's where things get hard. We now have 3 files, represented by 3 different directories, and only two fields to store them in (`source_path` and `generated_path`). The only way to fix this would be to change `DirectoryInfo` to have an extra field `genfiles_path`, which seems like a bad idea, given that it's only relevant for blaze, and blaze is getting rid of it soon.

rules/directory/glob.bzl
matts1: If this is going to be contentious, I'll split this into two PRs and we can can continue discussions on globbing in the other PR, since this is a hard blocker for the rules_cc work.

rules/directory/filter.bzl
```diff
@@ -0,0 +1,65 @@
...
+        get_relative(directory, path, require_file = True)
+        for path in ctx.attr.data
+    ]
+
+    return DefaultInfo(
+        files = srcs,
+        runfiles = ctx.runfiles(files = data, transitive_files = srcs),
+    )
+
+directory_filter = rule(
```
comius, line 35: Is it possible to implement directory_fileter using the glob rule in the other PR?

matts1: Strictly speaking, this rule isn't necessary - we could just tell a user to use a glob all the time (and the performance should be equivalent - globs are trivially cheap if we don't have a `*` in them). The main reason I didn't was just because if you had `*` in the filename it'd fail, and on principle it seemed weird to require a glob.

I don't have particularly strong feelings about this, so I'm happy to change this if you'd like me to. Let's discuss what to do with this in our meeting today.

rules/directory/directory.bzl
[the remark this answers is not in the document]
matts1: I think we have three options here:

## Explicitly store each path seperately
* + Works pretty well in bazel
* - Doesn't work great in blaze due to genfiles
* + Easy for a user to get both the source path and the generated path seperately
* Can use a helper function to get the path for a given directory

## Store them together (your `include_paths` idea)
* + Works well in both bazel and blaze
* - If a user does want specifically the source or generated path, it may be difficult
* Can use a helper function to get the path for a given directory.

## Only allow a single path
With this idea, if a user tries to put source files and generated files in the sources for a directory, we'll error out

* + Works well in both bazel and blaze
* + Easy to get the path
* - Can't store both source and generated files in a directory

Off the top of my head, I can't think of any situation where you'd want both source and generated files at the same time. I also think that if you want to do that, requiring a user to just have two separate rules seems pretty reasonable. Based on that, I think option 3 is the best, since it's the simplest to use. Instead of `include_paths`, we'd only have to store a single path.

rules/directory/filter.bzl
matts1: In the meantime, I've pulled this out of the CL.

tests/directory/BUILD
```diff
@@ -0,0 +1,104 @@
...
+)
+load(
+    ":subdirectory_test.bzl",
+    "nonexistent_subdirectory_test",
+    "subdirectory_as_file_test",
+    "subdirectory_of_file_test",
+    "subdirectory_test",
+)
+
+analysis_test(
```
comius, line 19: Having analysis_test in BUILD file, targets in test_data and assertions in directory_test.bzl file makes the tests hard to follow. Arrange, act and assert parts are in different locations

The idea behind analysis_test was that you can create all of the in the same file. That way the tests should be easier to follow. See: https://rules-testing.readthedocs.io/en/latest/analysis_tests.html

comius (changes requested):

.ijwb/.blaze/qsyncdata.gz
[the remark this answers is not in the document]
matts1: Sorry, I blew up my global .gitignore the other day, and it appears that this got in before I had the chance to fix it.

tests/directory/directory_test.bzl
```diff
@@ -0,0 +1,158 @@
...
+    testdata = directory_subject(env, root.actual.entries["testdata"])
+    testdata.entries().keys().contains_exactly(["f1", "subdir"])
+    testdata.human_readable().equals(human_readable + "/testdata")
+
+    subdir = directory_subject(env, testdata.actual.entries["subdir"])
+    subdir.entries().contains_exactly({"f2": f2})
+    subdir.transitive_files().contains_exactly([f2])
+    env.expect.that_str(subdir.actual.path + "/f2").equals(f2.path)
+
+def _source_root_test(name):
```
comius, line 48: Invert the order. First a test function, then impl. This way you can follow arrange, act, assert parts; reading it from top to bottom.

comius (approved):

comius (approved):

comius (approved):

PR #514, opened 2024-05-29
Add is_normalized and starts_with to paths module

comius:
Those two functions are often used by built-in rules. This will make it easier to move them out of Bazel binary.

### DIFF
--- a/docs/paths_doc.md
+++ b/docs/paths_doc.md
@@ -91,7 +91,7 @@ Returns `True` if `path` is an absolute path.
 paths.is_normalized(<a href="#paths.is_normalized-str">str</a>, <a href="#paths.is_normalized-look_for_same_level_references">look_for_same_level_references</a>)
 </pre>
 
-Returns true if the passed path contains uplevel references "..".
+Returns true if the passed path doesn't contain uplevel references "..".
 
 Also checks for single-dot references "." if look_for_same_level_references
 is `True.`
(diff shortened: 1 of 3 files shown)


lib/paths.bzl
```diff
@@ -153,6 +153,67 @@ def _normalize(path):
 
     return path or "."
 
+_BASE = 0
+_SEPARATOR = 1
+_DOT = 2
+_DOTDOT = 3
+
+def _is_normalized(str, look_for_same_level_references = True):
+    """Returns true if the passed path contains uplevel references "..".
```
hvadehra, line 162: contains -> does not contain

docs/paths_doc.md
```diff
@@ -83,6 +83,33 @@ Returns `True` if `path` is an absolute path.
...
 
+<a id="paths.is_normalized"></a>
+
+## paths.is_normalized
+
+<pre>
+paths.is_normalized(<a href="#paths.is_normalized-str">str</a>, <a href="#paths.is_normalized-look_for_same_level_references">look_for_same_level_references</a>)
+</pre>
+
+Returns true if the passed path contains uplevel references "..".
```
hvadehra, line 94: Are the docs autogenerated from code? If not, this and the RETURNS section below need changing.

lib/paths.bzl
```diff
@@ -153,6 +153,67 @@ def _normalize(path):
...
+        look_for_same_level_references: If True checks if path contains uplevel
+          references ".." or single-dot references ".".
+
+    Returns:
+        True if the path is normalized, False otherwise.
+    """
+    state = _SEPARATOR
+    for c in str.elems():
+        is_separator = False
+        if c == "/":
```
hvadehra, line 178: Possible incorrect for windows. The native code check is: https://github.com/bazelbuild/bazel/blob/34e3159e342a58b66353cca57ebe0b3d3ad85ff0/src/main/java/com/google/devtools/build/lib/vfs/WindowsOsPathPolicy.java#L231

comius: The native check actually uses the host operating system: https://github.com/bazelbuild/bazel/blob/34e3159e342a58b66353cca57ebe0b3d3ad85ff0/src/main/java/com/google/devtools/build/lib/vfs/PathFragment.java#L728

Since we're using `/` throughout the rules, even when they run on Windows, I believe this code is better than the native one. One piece of evidence are also all other functions here, that use just forward slashes.

docs/paths_doc.md
comius: Regenerated.

hvadehra (approved):

PR #518, opened 2024-05-31
Add Publish to BCR setup

comius:


.bcr/metadata.template.json
```diff
@@ -0,0 +1,9 @@
+{
+  "homepage": "https://github.com/bazelbuild/bazel-skylib",
+  "maintainers": [],
```
hvadehra, line 3: I was informed on a recent rules_java release that the BCR would like all modules to have a list of maintainers here. Maybe add some of the owners of this repo?

hvadehra (approved):

.bcr/metadata.template.json
comius: Done.

PR #522, opened 2024-06-06
Strip compatible_with from bzl_library rules

aiuto:
Wrap bzl_library in a macro so that we can force off some global attributes that are never appropriate for analysis of build rules.

The motivation is that when people use `package(default_compatible_with=X)` alongside `bzl_library`, they have accidentally marked their _build rules_ as eligible to deliver to some target environment. This can cause `...` scoped tests on cross platform projects to fail because the user defined rules are not the code we are shipping.

An alternative idea was to use a rule initializer for this. That is the better solution, but it would require a newer version of Bazel, so I opted for the backwards compatible solution.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -60,6 +60,9 @@ bzl_library(
 bzl_library(
     name = "bzl_library",
     srcs = ["bzl_library.bzl"],
+    deps = [
+        "//rules/private:bzl_library",
+    ],
 )
 
 bzl_library(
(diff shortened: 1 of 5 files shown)


aiuto:
cc: @ murali42

aiuto:
cc: @gregestren

brandjon:
Er, didn't mean to remove other reviewers per se, just meant I'd like to ensure Sasha makes the decision here since this concerns bzl_library.

gregestren:
Looks reasonable to me.

comius (approved):

comius:
@aiuto should we merge this?

aiuto:
I think we should merge it.  I no longer have merge rights, so I didn't do it myself.

PR #525, opened 2024-06-24
Add support for data in expand_template

keith:
This can be useful for encoding the paths to things in your expansions


rules/expand_template.bzl
```diff
@@ -16,10 +16,21 @@
 """
 
 def _expand_template_impl(ctx):
+    expanded_substitutions = {}
+    for key, value in ctx.attr.substitutions.items():
+        expanded_substitutions[key] = ctx.expand_make_variables(
```
comius, line 21: ctx.expand_make_variable is marked as "Deprecated" in Bazel documentation. People before me marked it that way and I think we shouldn't introduce new uses.

Please remove it's use.

I believe it's implementable in Starlark, and skylib would be a nice location for such an implementation. (And we might already have one implementation in the C++ rules in builtins).

The rest of the PR seems fine.

tests/expand_template/BUILD
```diff
@@ -14,21 +14,36 @@
...
+    ],
     substitutions = {
         "@name@": "test",
         "@version@": "1.1.1",
+        "@path@": "$(rlocationpath :version)",
+        "@toolchain@": "prefix$(SOMEVAR)suffix",
     },
     template = "test.tpl.yaml",
+    toolchains = [
+        ":my_string_flag",
```
comius, line 45: I didn't realise we can commit such attrocities as putting a flag under toolchains. I've got mixed feeling though, becuase it looks like a nice and elegant interface.

comius (changes requested):

tests/expand_template/BUILD
keith: yea, idk if it's great or not but at least in this case it means we don't have to introduce another custom thing just for tests

rules/expand_template.bzl
keith: relevant convo https://groups.google.com/g/bazel-discuss/c/WEnG-WocaTc

so you're saying we should make a new function to do the same thing in skylib and use that here?

comius: Yes. I bumped https://github.com/bazelbuild/bazel/issues/5859 to P3.

And I noticed also https://github.com/bazelbuild/bazel-skylib/pull/486 which might already be doing that. Would you like to help in review?

There might be other PRs, I haven't gone through all of them yet.

PR #527, opened 2024-06-29
feat: case insensitive diff_test

peakschris:
In my work on windows quality in rulesets, a frequent issue I am finding is that diff_tests are failing due to differences in line endings -- either due to differences in the 'expected' values, or cross-platform differences in generators (e.g. buildifier always writes LF, whilst skydoc writes LF or CRLF depending on platform).

We can't expect rule authors to work around these differences, because in practice they do not have good access to windows dev hardware.

The goal of this PR is to make diff_test slightly less brittle and ignore line endings by default. An `ignore_line_endings` flag is provided to switch this behavior.

This flag is supported by the `diff` command on unix. On windows, it requires the `tr.exe` command that is part of most bash installs including msys2. If `tr.exe` cannot be found the `ignore_line_endings` flag is silently ignored and set to False.

This PR allows us to enable the 1 disabled test 'stardoc_with_diff_test' on windows CI, and also fixes many test failures in dependent rulesets such as rules_starlib and rules_bazel_integration_test

### DIFF
--- a/rules/diff_test.bzl
+++ b/rules/diff_test.bzl
@@ -35,8 +35,22 @@ def _diff_test_impl(ctx):
 @echo off
 SETLOCAL ENABLEEXTENSIONS
 SETLOCAL ENABLEDELAYEDEXPANSION
-set MF=%RUNFILES_MANIFEST_FILE:/=\\%
 set PATH=%SYSTEMROOT%\\system32
+if defined RUNFILES_MANIFEST_FILE (
+    set MF=%RUNFILES_MANIFEST_FILE:/=\\%
+) else (
+    if exist MANIFEST (
+        set MF=MANIFEST
+    ) else (
+        if exist ..\\MANIFEST (
+            set MF=..\\MANIFEST
+        )
+    )
+)
+if not exist %MF% (
+    echo Manifest file %MF% not found
+    exit /b 1
+)
 set F1={file1}
 set F2={file2}
 if "!F1:~0,9!" equ "external/" (set F1=!F1:~9!) else (set F1=!TEST_WORKSPACE!/!F1!)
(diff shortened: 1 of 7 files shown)


peakschris:
This fixes:
- https://github.com/bazelbuild/bazel-skylib/issues/528
- https://github.com/bazelbuild/bazel-skylib/issues/529

peakschris:
@brandjon @tetromino @comius @hvadehra @c-mita gentle ping, would it be possible for someone to review this? I have a chain of PRs to rulesets dependent on this one to improve windows support.

cgrindel:
@peakschris Did you see [this message](https://groups.google.com/g/bazel-discuss/c/dFhhzJpVggg/m/h1cpy6UWAQAJ)? This may not be reviewed and merged. I will bring this up at the next Community for Bazel technical steering committee meeting which is scheduled for Tuesday, July 9.

rules/diff_test.bzl
```diff
@@ -79,10 +101,36 @@ if "!RF2!" equ "" (
...
+rem todo: in future better to pull in diff.exe to align with non-windows path
+if "{ignore_line_endings}"=="1" (
+  if exist {bash_bin} (
+    for %%f in ({bash_bin}) do set "TR=%%~dpf\\tr.exe"
+  ) else (
+    rem match bazel's algorithm to find unix tools
+    set "TR=C:\\msys64\\usr\\bin\\tr.exe"
+  )
+  if not exist !TR! (
+    echo>&2 WARNING: ignore_line_endings set but !TR! not found; line endings will be compared
```
comius, line 114: Bazel team is quite opinionated about warnings. This should be an error.

rules/diff_test.bzl
```diff
@@ -26,17 +26,39 @@ def _runfiles_path(f):
...
 
+def _ignore_line_endings(ctx):
+    ignore_line_endings = "0"
+    if ctx.attr.ignore_line_endings:
+        ignore_line_endings = "1"
+    return ignore_line_endings
+
 def _diff_test_impl(ctx):
     if ctx.attr.is_windows:
+        bash_bin = ctx.toolchains["@bazel_tools//tools/sh:toolchain_type"].path
```
comius, line 37: Now that we require bash, I wonder if the maintenance of this rule would be simpler, to use the same bash script on windows (and maybe fallback to fc or just use diff)?

tests/diff_test/diff_test_tests.sh
```diff
@@ -80,12 +80,13 @@ eof
   echo bar > "$ws/$subdir/b.txt"
 
   (cd "$ws" && \
-   bazel test ${flags} "//${subdir%/}:same" --test_output=errors 1>"$TEST_log" 2>&1 \
+   bazel test ${flags} "//${subdir%/}:same" --test_output=errors --noshow_progress 1>>"$TEST_log" 2>&1 \
```
comius, line 83: Why are you changing tests?

rules/diff_test.bzl
```diff
@@ -120,14 +170,19 @@ else
   echo >&2 "ERROR: could not find \"{file1}\" and \"{file2}\""
   exit 1
 fi
-if ! diff "$RF1" "$RF2"; then
-  echo >&2 "FAIL: files \"{file1}\" and \"{file2}\" differ. "{fail_msg}
+if ! diff {strip_trailing_cr}"$RF1" "$RF2"; then
+  MSG={fail_msg}
+  if [[ "${{MSG}}" == "" ]]; then
+      MSG="why? diff {strip_trailing_cr}"${{RF1}}" "${{RF2}}" | cat -v"
```
comius, line 176: Why is this message so different from the message before? Also use of RF1 is inconsistend with use of {file1} below.

comius (changes requested):
I think this is quite complicated change and this brings with it potentially higher mainteinance costs.

PR #531, opened 2024-07-12
fix: subpackages.all works for root packages

kilian-funk:
This fixes "subpackages.all" which delivers broken results in case it is used in the root package.

Assuming there is a package `//foo`. Then `subpackages.all` returns `["///foo"]`. With this fix it return the expected `["//foo"]`.

### DIFF
--- a/lib/subpackages.bzl
+++ b/lib/subpackages.bzl
@@ -62,7 +62,10 @@ def _all(exclude = [], allow_empty = False, fully_qualified = True):
     return subs
 
 def _fully_qualified(relative_path):
-    return "//%s/%s" % (native.package_name(), relative_path)
+    package_name = native.package_name()
+    if package_name:
+        return "//%s/%s" % (package_name, relative_path)
+    return "//" + relative_path
 
 def _exists(relative_path):
     """Checks to see if relative_path is a direct subpackage of the current package.


lib/subpackages.bzl
```diff
@@ -62,7 +62,10 @@ def _all(exclude = [], allow_empty = False, fully_qualified = True):
     return subs
 
 def _fully_qualified(relative_path):
-    return "//%s/%s" % (native.package_name(), relative_path)
+    package_name = (
+        "%s/" % native.package_name() if native.package_name() else ""
```
comius, line 66: This code is a bit hard to understand. On first read I though it was 
`"%s/" % (native.package_name() if native.package_name() else "")`

Consider something simpler like:
```
if native.package_name():
  return "//" + relative_path
else:
  return "//%s%s" % (package_name, relative_path)
```

comius (changes requested):

lib/subpackages.bzl
comius: Ups, swapped then and else branch.

kilian-funk: @comius Thanks for your feedback. I changed the code. It should be better readable now.

comius (approved):

PR #544, opened 2024-10-08
README.md: fix broken directory rule links

lummax:


README.md
```diff
@@ -63,8 +63,8 @@ s = shell.quote(p)
 * [common_settings](docs/common_settings_doc.md)
 * [directories](docs/copy_directory_doc.md)
     * [directory](docs/directory_doc.md)
-    * [directory_glob](docs/directory_glob.md)
-    * [subdirectory](docs/subdirectory_doc.md)
+    * [directory_glob](docs/directory_glob_doc.md)
+    * [directory_subdirectory](docs/directory_subdirectory_doc.md)
```
brandjon: Don't change the name from `subdirectory` to `directory_subdirectory`. Apparently, `directory_subdirectory_doc.md` is the name of the documentation page but not of the rule itself (see the .bzl file and the title of the md page).

README.md
```diff
@@ -63,8 +63,8 @@ s = shell.quote(p)
 * [common_settings](docs/common_settings_doc.md)
 * [directories](docs/copy_directory_doc.md)
```
brandjon, line 64: Can you also fix this to not be a link? `copy_directory_doc.md` is clearly an error and I don't see any page corresponding to this node.

brandjon (commented):
Thanks! Requested a couple fixes.

README.md
lummax: Yes, I have already pushed an update commit.

README.md
lummax: This would be https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_directory_doc.md

README.md
brandjon: Whoops, thanks.

README.md
brandjon: Right, that's linked to below in the bullet point on line 68. But on line 64, the parent item for these sub-bullets, it doesn't seem like `directories` should be a link to anything.

lummax: Right, done.

brandjon (approved):

PR #549, opened 2024-10-22
Bump `rules_license` to 1.0.0

shs96c:
This brings in the `purl` field that is increasingly used by other libraries. In the `bzlmod` world this is less important, but for folks using workspace-based builds, skylib tends to be loaded early so any deps it brings in normally end up being used.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -11,7 +11,7 @@ register_toolchains(
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
-bazel_dep(name = "rules_license", version = "0.0.7")
+bazel_dep(name = "rules_license", version = "1.0.0")
 
 ### INTERNAL ONLY - lines after this are not included in the release packaging.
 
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -11,7 +11,7 @@ register_toolchains(
 )
 
 bazel_dep(name = "platforms", version = "0.0.4")
-bazel_dep(name = "rules_license", version = "0.0.7")
+bazel_dep(name = "rules_license", version = "1.0.0")
```
brentleyjones, line 14: Dow e want to _force_ 1.0.0 on people? I'm only asking because `MODULE.bazel` should declare the minimum version needed.

shs96c: It should be fine to update everyone to use 1.0.0: it's not hugely different from the previous release and I don't think it introduces new dependencies. OTOH, it includes the `purl` attribute in the `PackageInfo` which is useful when generating SBOMs

PR #550, opened 2024-10-22
Remove sh_binary usage for Bazel 8 friendliness

tetromino:
We can use our own native_binary instead


tetromino:
@aranguyen - the gazelle CI failure on Bazel 8 is caused by https://github.com/bazel-contrib/bazel-gazelle/issues/1959

aranguyen (approved):

PR #551, opened 2024-10-25
Add a repeatable string list flag rule to skylib

nreid260:


brentleyjones:
**rules_swift** would migrate to this: https://github.com/bazelbuild/rules_swift/blob/eb60754365ab3cf9ad2328bfd63803608188da31/swift/internal/build_settings.bzl#L19-L34

comius (changes requested):
LGTM, please also wire up docs in https://github.com/bazelbuild/bazel-skylib/blob/main/docs/common_settings_doc.md

nreid260:
Done

PR #553, opened 2024-11-15
Allow analysis tests to support `for_dependency_resolution` on the `target_under_test`

Bocete:
This attribute is necessary to analysis test dormant targets. It's hard to assert anything useful without materializing what's behind the `target_under_test`, and that requires setting `for_dependency_resolution` on the `target_under_test`.

See https://bazel.build/rules/lib/toplevel/attr#label.for_dependency_resolution for more info on this attribute. It's a new bazel feature.

### DIFF
--- a/lib/unittest.bzl
+++ b/lib/unittest.bzl
@@ -227,6 +227,7 @@ def _make_analysis_test(
         fragments = [],
         config_settings = {},
         extra_target_under_test_aspects = [],
+        target_under_test_for_dependency_resolution = False,
         doc = ""):
     """Creates an analysis test rule from its implementation function.
 
@@ -267,6 +268,9 @@ def _make_analysis_test(
           flags in a single build
       extra_target_under_test_aspects: An optional list of aspects to apply to the target_under_test
           in addition to those set up by default for the test harness itself.
+      target_under_test_for_dependency_resolution: If true, the target_under_test will be used for
+          dependency resolution. See
+          https://bazel.build/rules/lib/toplevel/attr#label.for_dependency_resolution for more info.
       doc: A description of the rule that can be extracted by documentation generating tools.
 
     Returns:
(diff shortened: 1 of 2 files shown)


Bocete:
I suspect this is blocked on a blaze release? Dormant deps are still experimental, and the CI tests are failing because of syntax that's made available via experimental flags. @lberki to comment

lberki:
I don't think it's technically blocked on a Blaze release, only its use internally is for reasons complicated and irrelevant to Bazel. So as long as the code is fine, feel free to merge it.

Bocete:
Hmm the build checks, here on github, are failing because:

```
(23:17:46) ERROR: Traceback (most recent call last):
	File "/workdir/tests/unittest_tests.bzl", line 284, column 28, in <toplevel>
		"dormant_dep": attr.dormant_label(),
Error: 'attr' value has no field or method 'dormant_label'
```

That's telling me something needs to be enabled blaze-side, no? Or maybe the CI system here needs to be tweaked to add a flag? Please tell me what to do here, thanks!

lberki:
I'm not familiar with the unit testing framework so take the reviewers' advice over mine, but how about adding something like this:

```
if hasattr(attr, "dormant_label"):
  inspect_dormant_target_fake_rule = rule(attrs = {"dormant_dep": attr.dormant_label()})
else:
  inspect_dormant_target_fake_rule = rule(implementation = _always_fail)
```

that way, Skylib remains compatible with older Bazel versions while still allowing you to add these symbols.

Bocete:
Hmm I don't think I can do that, `attr` belongs to `ctx` and that's all analysis phase stuff. Unless I'm missing some way to know if dormant deps are supported in the loading phase?

I'll defer to the reviewers too. I don't know how to submit tests to this until all Blaze versions support dormant deps, but at the same time, I need the analysis testing framework to support them. The implementation could be submitted fine, it doesn't actually use any of the new syntax, it's the tests of the implementation that wouldn't work.

Bocete:
Oh I was looking at `attrs`, not `attr`. Let me try that alternative

brandjon:
I think @comius would be the applicable reviewer, but didn't the source of truth get moved to rules_testing? Or is that not effective yet?

comius:
@Bocete The change looks good. It might be easier to move it into rules_testing. You can make it internally (the copybara setup is bidirectional). 

In bazel-skylib, it would be impossible to test it, because the tests are running against released Bazel version, so there's no version with dormant deps.

In rules_testing, you can do internal tests with Blaze, and disable them on the export until we get Bazel 8.

PR #561, opened 2025-03-12
Add new "env" variable to native_* rules

gioachin:
This (standard) variable "env" allows the propagation of environment variables to the test being executed or tested.


gioachin:
Hi, would it be possible to get this patch reviewed? (@hvadehra ?)
Thanks

hvadehra (approved):

PR #571, opened 2025-05-12
feat: generate doc_extract targets

alexeagle:
Allows bzl_library targets to produce documentation without needing to load stardoc or add additional targets. Note, it would be better to do this with an aspect that visits the bzl_library graph, but the starlark_doc_extract rule is in Bazel java core, and so we cannot spawn actions using its Java code. It needs a main() entry point.

### DIFF
--- a/bzl_library.bzl
+++ b/bzl_library.bzl
@@ -23,7 +23,7 @@ load(
 
 StarlarkLibraryInfo = _StarlarkLibraryInfo
 
-def bzl_library(name, **kwargs):
+def bzl_library(name, srcs = [], deps = [], **kwargs):
     """Wrapper for bzl_library.
 
     Args:
@@ -36,11 +36,21 @@ def bzl_library(name, **kwargs):
     _ = kwargs.pop("exec_compatible_with", None)
     _ = kwargs.pop("features", None)
     _ = kwargs.pop("target_compatible_with", None)
+    
     _bzl_library(
         name = name,
+        srcs = srcs,
+        deps = deps,
         compatible_with = [],
         exec_compatible_with = [],
         features = [],
         target_compatible_with = [],
         **kwargs
     )
+    
+    for src in srcs:
+        native.starlark_doc_extract(
+            name = "{}.doc_extract{}".format(name, "_" + src if len(srcs) > 1 else ""),
+            src = src,
+            deps = deps,
+        )


comius (commented):
A PR like this would create a lot of additional targets. There's a cost connected with each additional target. Such feature would need to be disabled at Google.

A lot of automatically created targets are potentially unused. Usually to generate documentation of one rules_* repo, you need a couple of starlar_doc_extract rules, while the number of bzl_libraries is much larger.

alexeagle:
What do you think of the alternative: making it possible to write an aspect that invokes starlark_doc_extract actions? If someone worked on such a PR to bazelbuild/bazel do you think it might be accepted?

bzl_library.bzl
```diff
@@ -38,9 +38,22 @@ def bzl_library(name, **kwargs):
...
         exec_compatible_with = [],
         features = [],
         target_compatible_with = [],
         **kwargs
     )
+    enable_doc_extract = select({
+        Label("//:extract_docs_flag"): hasattr(native, "starlark_doc_extract"),
+        "//conditions:default": False,
+    })
+    if enable_doc_extract:
```
comius, line 53: selects don't work in a macro

comius:
> What do you think of the alternative: making it possible to write an aspect that invokes starlark_doc_extract actions? If someone worked on such a PR to bazelbuild/bazel do you think it might be accepted?

I don't fully understand how to make such an aspect possible without incurring tech-debt.

What's the intent? Automatically extracting all the docs? I think we have some support in query to do that. cc @tetromino

alexeagle:
> What's the intent?

Yes I'm experimenting with a pipeline to generate docs, similar to readthedocs. @rickeylev and I discussed in a rules authors SIG meeting. https://github.com/alexeagle/doc.bzl is the prototype.

alexeagle:
Another intent here is merely to ensure that rulesets have correct `bzl_library` targets.

As I'm visiting the 720 modules on BCR I find that many are missing deps, as one example https://github.com/aspect-build/rules_swc/pull/305

You might say those repos ought to include more testing to avoid such a defect, but this is extra burden on those maintainers. It's akin to `proto_library` having no descriptor actions and therefore no validation of the `deps` until some `lang_proto_library` rule depends on it.

rickeylev:
Another approach would be to let starlark_doc_extract take a bzl_library as input? Or multiple files as inputs.

If memory serves, it isn't possible today because of the combination of bzl_library not returning a single output, and starlark_doc_extract requiring a single_file_only input. In rules_python, I worked around this by having a custom rule take a bzl_library as input and produce a single output to then feed into starlark_doc_extract, and just forcing the convention of every bzl_library having a single srcs.

PR #574, opened 2025-05-30
Migrate away from `@bazel_tools//src/conditions:host_windows`

fmeum:
The condition has been deprecated at HEAD. None of its usages in skylib were correct as they should all either match on the target or the exec platform, not the host.


meteorcloudy (approved):
Thank you, approve from Windows perspective!

fmeum:
Friendly ping, would be great to have this merged and released

PR #576, opened 2025-06-03
Use `native.package_relative_label` in `selects.bzl`

fmeum:
Prepares for a future in which `select` may eagerly resolve string keys to `Label`s.


fmeum:
@Wyverald

fmeum:
@tetromino Would be great to have this merged to unblock https://github.com/bazelbuild/bazel/issues/26281.

Wyverald (approved):

PR #578, opened 2025-06-13
Add `structs.merge`

mattyclarkson:
```python
one = struct(a = 1)
two = struct(b = 2)
overwrite = struct(a = 3)
merged = structs.merge(one, two , overwrite)
# struct(a = 3, b = 2)
```

### DIFF
--- a/lib/structs.bzl
+++ b/lib/structs.bzl
@@ -14,6 +14,20 @@
 
 """Skylib module containing functions that operate on structs."""
 
+_built_in_function = type(str)
+
+def _is_built_in_function(v):
+    """Returns True if v is an instance of a built-in function.
+
+    Args:
+      v: The value whose type should be checked.
+
+    Returns:
+      True if v is an instance of a built-in function, False otherwise.
+    """
+
+    return type(v) == _built_in_function
+
 def _to_dict(s):
     """Converts a `struct` to a `dict`.
 
@@ -31,7 +45,7 @@ def _to_dict(s):
     return {
         key: getattr(s, key)
         for key in dir(s)
-        if key != "to_json" and key != "to_proto"
+        if not ((key == "to_json" or key == "to_proto") and _is_built_in_function(getattr(s, key)))
     }
 
 def _merge(first, *rest):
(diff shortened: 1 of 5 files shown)


lib/structs.bzl
```diff
@@ -34,6 +34,22 @@ def _to_dict(s):
         if key != "to_json" and key != "to_proto"
```
fmeum, line 34: This method looks suspect: These functions no longer exist in the latest version of Bazel, but they may very well be used for other purposes and should be copied in that case.

Before adding a new function, could we make this more precise by checking the `type` of the values before skipping over these keys?

mattyclarkson: > These functions no longer exist in the latest version of Bazel

They do not:

```
                print(struct().to_json())
Error: 'struct' value has no field or method 'to_json'
```

However, I cannot create a struct to test:

```
                structs.to_dict(struct(to_json = 1, to_proto = True)),
Error in struct: cannot override built-in struct function 'to_json'
```

I can implement the `type` check but won't be able to add a test.

fmeum: You can test with `last_green`: https://github.com/bazelbuild/bazel/commit/db5906870bdcb1631e1b18be8637aa21b79cf943

fmeum: https://github.com/bazel-contrib/bazel_features/pull/103

mattyclarkson:
I am unclear why this is failing the `buildifier` job in the CI.

`buildifier` (same version as CI) is clean for me locally on `lib/structs.bzl`, `tests/structs_test.bzl`, `MODULE.bazel` and `WORKSPACE` which have been modified.

Maybe an infra-failure?

mattyclarkson:
@fmeum this is good to go.

fmeum:
@mattyclarkson I don't have write access, I can just help you get this in your shape for review by Google folks

lib/structs.bzl
```diff
@@ -31,9 +45,25 @@ def _to_dict(s):
     return {
         key: getattr(s, key)
         for key in dir(s)
-        if key != "to_json" and key != "to_proto"
+        if not ((key == "to_json" or key == "to_proto") and _is_built_in_function(getattr(s, key)))
```
eric-wieser, line 48: This looks like an unrelated change to me; perhaps it should be in its own PR?

PR #581, opened 2025-06-30
Add BCR publishing workflow

fmeum:
This still requires a PAT to be added as BCR_PUBLISH_TOKEN.


.bcr/gazelle/presubmit.yml
```diff
@@ -1,13 +1,17 @@
...
-  - ubuntu2004
+  - macos_arm64
+  - ubuntu2204
+  - ubuntu2404
   - windows
   bazel:
+  - 8.x
   - 7.x
   - 6.x
+  - 5.x
```
meteorcloudy, line 14: I think we can remove `5.x` since it's already deprecated.

meteorcloudy (changes requested):
Nice!

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,35 @@
...
+        description: git tag being released
+        required: true
+        type: string
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.2.1
+    with:
+      tag_name: ${{ inputs.tag_name }}
+      # GitHub repository which is a fork of the upstream where the Pull Request will be opened.
+      registry_fork: bazel-contrib/bazel-central-registry
```
meteorcloudy, line 27: Hmm, does this mean I actually need a token that has access to this fork? I will probably use the `bazel-io` account, so maybe this should be `bazel-io/bazel-central-registry`?

/cc @alexeagle @kormide

alexeagle: Yes the token has to be able to push to a fork that the PR may originate from

meteorcloudy:
I've setup BCR_PUBLISH_TOKEN using the `bazel-io` account for bazelbuild org, if this is successful, we can apply the same release process to rules_java, rules_cc, etc..

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,35 @@
...
+  # the workflow from the GitHub UI
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: git tag being released
+        required: true
+        type: string
+jobs:
+  publish:
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v0.2.1
```
kormide, line 23: Latest version is `0.2.2`.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,35 @@
...
+    with:
+      tag_name: ${{ inputs.tag_name }}
+      # GitHub repository which is a fork of the upstream where the Pull Request will be opened.
+      registry_fork: bazel-io/bazel-central-registry
+      draft: false
```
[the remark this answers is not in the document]
meteorcloudy: Yeah, it doesn't matter.

PR #587, opened 2025-07-14
Adding allowed values for scope attribute

aranguyen:
Adding allowed values for scope attribute

### DIFF
--- a/rules/common_settings.bzl
+++ b/rules/common_settings.bzl
@@ -69,6 +69,7 @@ int_flag = rule(
         "scope": attr.string(
             doc = "The scope indicates where a flag can propagate to",
             default = "universal",
+            values = ["universal", "project", "exec"],
         ),
     },
     doc = "An int-typed build setting that can be set on the command line",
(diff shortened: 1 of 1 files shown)


tetromino (changes requested):
There is a lot of unnecessary repetiton here. I would suggest:

1. Moving the scope attr into a variable (like _MAKE_VARIABLE_ATTR already is), so it's defined in one place and can be used by all the rules;
2. Adding documentation for what the enum values mean.

So something like:

```
_SCOPE_ATTR = attr.string(
    doc = """The scope indicates where a flag can propagate to.

Possible values:
* `"universal"` - means ...
...
""",
    default = "universal",
    values = ["universal", "project", "exec"],
)

...

int_flag = rule(
    ...
    attrs = {
        "make_variable": _MAKE_VARIABLE_ATTR,
        "scope": _SCOPE_ATTR,
    },
    doc = "An int-typed build setting that can be set on the command line",
)
```

PR #588, opened 2025-07-15
Migrate `copy_directory` away from deprecated host constraint

fmeum:
This was missed in #574


fmeum:
@meteorcloudy @sluongng

meteorcloudy (approved):

meteorcloudy:
How urgent is this? Do we need a release now or could we wait a bit longer?

PR #593, opened 2025-09-17
Introduce repeatable_string_flag.

susinmotion:
New settings of this flag type do not override each other; they are concatenated into a list. --foo=bar --foo=baz -> [bar, baz]. This is the equivalent of a string typed native flag with allow_multiple.

### DIFF
--- a/rules/common_settings.bzl
+++ b/rules/common_settings.bzl
@@ -99,7 +99,7 @@ bool_setting = rule(
     implementation = _impl,
     build_setting = config.bool(),
     attrs = {
-	"scope": _SCOPE_ATTR,
+        "scope": _SCOPE_ATTR,
     },
     doc = "A bool-typed build setting that cannot be set on the command line",
 )
@@ -117,13 +117,13 @@ def _repeatable_string_flag_impl(ctx):
     return BuildSettingInfo(value = ctx.build_setting_value)
 
 repeatable_string_flag = rule(
-    implementation = _flag_impl,
+    implementation = _repeatable_string_flag_impl,
     build_setting = config.string_list(
         flag = True,
         repeatable = True,
     ),
     attrs = {
-	"scope": _SCOPE_ATTR,
+        "scope": _SCOPE_ATTR,
     },
     doc = "A string-typed build setting that can be set on the command line. Multiple settings do not overwrite each other; they are concatenated into a list",
 )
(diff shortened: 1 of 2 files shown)


rules/common_settings.bzl
```diff
@@ -123,6 +123,24 @@ string_list_flag = rule(
     doc = "A string list-typed build setting that can be set on the command line",
 )
 
+def _flag_impl(ctx):
```
tetromino, line 126: Nit: rename to `_repeatable_string_flag_impl` - there are a lot of flag rules in this file!

rules/common_settings.bzl
```diff
@@ -123,6 +123,24 @@ string_list_flag = rule(
...
+    implementation = _flag_impl,
+    build_setting = config.string_list(
+        flag = True,
+        repeatable = True,
+    ),
+    attrs = {
+        "scope": attr.string(
+            doc = "The scope indicates where a flag can propagate to",
+            default = "universal",
+        ),
```
tetromino, line 139: Pre-existing issue: we should move scope to a variable (like `_MAKE_VARIABLE_ATTR`) to avoid so much repetition...

Can be done in follow-up

rules/common_settings.bzl
```diff
@@ -123,6 +123,24 @@ string_list_flag = rule(
...
+    },
+    doc = "A string-typed build setting that can be set on the command line. Multiple settings do not overwrite each other; they are concatenated into a list",
```
tetromino, line 141: Please clarify in the docs here and for `string_list_flag` how the two rules differ. For example,

* for `string_list_flag` - "A string list-typed build setting which expects its value on the command line to be given in comma-separated format; for example, `--//my/setting=foo,bar` will be parsed as `["foo", "bar"]`. Contrast with `repeatable_string_flag`.
* for `repeastable_string_flag` - "A string-typed build setting which can be specified multiple times on the command line, with the values concatenated into a single list; for example, `--//my/setting=foo --//my/setting=bar` will be parsed as `["foo", "bar"]`. Contrast with `string_list_flag`.

tetromino (approved):
Overall lgtm

Nit: is it worth adding a test for this? (See tests/common_settings/BUILD)

rules/common_settings.bzl
susinmotion: Great doc suggestions!

susinmotion:
> Overall lgtm
> 
> Nit: is it worth adding a test for this? (See tests/common_settings/BUILD)

For posterity, I did not add a test because this flag type doesn't become a variable, so it's not a possible to make the comparable test. I tried using a custom starlark rule that had the an attr of this flag type, but it's overkill.

PR #605, opened 2025-12-19
Revert allow_symlink = True by default for non-executables

### LINKED ISSUE #604: Regression in 1.9.0: copy_file sets allow_symlink = True by default
Symlinks are not transparent to users.

For example, in the JS ecosystem, when importing a module, the real path of the module matters for resolution of the dependencies of the module itself.

Setting `allow_symlink = True` by default (in #565) breaks places where a module has been deliberately copied for the purpose of correct dependency resolution.

gzm0:
Partial revert of #565, fixes #604.


fmeum:
I guess this part of my change was overly optimistic - sorry for that! 

Now that Bazel has faster ways to copy, we could introduce a version of `ctx.actions.symlink` that does have all the performance optimizations (in particular no duplicate hashing), but still no discernable difference in semantics. I will look into that. 

Cc @tjgq

fmeum:
@tetromino @gregestren Could you cut a 1.9.1 release with this fix? Folks on slack kept hitting the regression since the fix hasn't been released.

PR #608, opened 2026-01-14
No-op: change written default scope from 'universal' to 'target'

susinmotion:
But this is ignored in favor of --incompatible_exclude_starlark_flags_from_exec_config. see https://github.com/bazelbuild/bazel/issues/26909


susinmotion:
I don't think we honor the scope unless the rule specifies it directly.

see: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java#L243-L244

and https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/skyframe/BuildOptionsScopeFunction.java#L214-L219

tetromino:
Aha, thanks, I missed the `isAttributeValueExplicitlySpecified` bit

PR #611, opened 2026-01-30
Change from stamp to some other boolean flag

justinhorvitz:
This removes a dependency on the value of `--stamp`, which we are using for internal caching.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -29,14 +29,6 @@ matrix:
   ? "@bazel_skylib_gazelle_plugin//..."
 
 tasks:
-  run_presubmit_bazel_6:
-    name: "Bazel 6 (WORKSPACE)"
-    platform: ${{ platform }}
-    bazel: 6.x
-    build_flags: *reusable_build_flags
-    test_flags: *reusable_test_flags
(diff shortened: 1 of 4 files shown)


justinhorvitz:
> The stamp check was originally added by @gregestren in [6a6a509](https://github.com/bazelbuild/bazel-skylib/commit/6a6a509f367ca0446e75c6f881719f5da2a55a26) - I'm wondering if there's a reason why it needed to be this way.

Pretty sure that stamp is just the shortest boolean option. Maybe we should define a bool starlark flag instead, instead of choosing an arbitrary native flag?

lib/BUILD
```diff
@@ -1,4 +1,6 @@
 load("//:bzl_library.bzl", "bzl_library")
+load("///rules:common_settings.bzl", "bool_setting")
```
gregestren: CI failing: from "///"?

gregestren:
> Wonder why CI was failing - unrelated pre-existing issue?

Bug fixed by https://github.com/bazelbuild/bazel-skylib/pull/611/changes/4c9a8f5aa929f197bc87ac2b32af41defa23b1b3.

PR #614, opened 2026-02-06
Adding type testing for Targets, Labels, and Files

alphalex-google:
While these types cannot be created in the loading phase (i.e. `value = type(File())`) for internally consistent type checking, these type rituals allow for a central definition of how/where this type checking can be governed, allowing for future changes, etc to all become generationally compatible as appropriate.

### DIFF
--- a/tests/types_tests.bzl
+++ b/tests/types_tests.bzl
@@ -17,6 +17,9 @@ load("//lib:new_sets.bzl", "sets")
 load("//lib:types.bzl", "types")
 load("//lib:unittest.bzl", "asserts", "unittest")
 
+# A placeholder target for unit tests that need a target.
+_target_test_hollow_target = "types_test_hollow_target"
+
 def _a_function():
     """A dummy function for testing."""
     pass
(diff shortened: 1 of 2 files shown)


tetromino (commented):
LGTM. I think CI is failing because docs need to be regenerated; let me take a look

tetromino (approved):

tests/types_tests.bzl
```diff
@@ -231,7 +234,92 @@ def _is_set_test(ctx):
...
+def _is_file_test(ctx):
+    """Unit test for types.is_file."""
+    env = unittest.begin(ctx)
+
+    # Ensuring a file does indeed match.
+    asserts.true(env, types.is_file(_make_test_file(ctx)))
+
+    asserts.false(env, types.is_file(99))
+    asserts.false(env, types.is_file(""))
+    asserts.false(env, types.is_file(set()))
```
tetromino, line 260: Here and everywhere below - please remove usage of `set()` because it requires Bazel 8.x (skylib is supposed to be compatible down to Bazel 7).

tetromino (changes requested):
@alphalex-google - please run `bazelisk run //docs:update` to regenerate docs/types_doc.md

PR #621, opened 2026-05-01
Fix gazelle not updating deps

keith:
Previously deps were set correctly when generating the first
bzl_library, but not when the deps changed overtime


keith:
cc @tetromino

PR #627, opened 2026-06-17
Fix bzl_library reordering

keith:
Previously depending on other context in the BUILD file this gazelle
plugin would move bzl_library targets even if they didn't change at all.


tetromino (changes requested):
Thank you for taking a look at this! The existing logic was obviously wrong in multiple ways.

However, for your proposed change, I don't understand why the canonical name matters: if all of a bzl_library's sources files have been deleted, shouldn't we delete the bzl_library target regardless of what it is named?

keith:
added another test case to clarify. it's about making sure we replace targets that have the wrong canonical name with the correct one, even when the file still exists

keith:
oh my new test case actually depends on https://github.com/bazelbuild/bazel-skylib/pull/621 as is, can you review that one first

PR #633, opened 2026-07-28
Fix default condition removal in `selects.bzl`

thii:
`_remove_default_condition` compared the whole settings list against
`"//conditions:default"` instead of comparing each element, so the entry
was never filtered out. Multi-element `config_setting_groups` that
include it then leak `"//conditions:default"` into the generated alias
chain, where it is not a real target, and analysis fails:

    selects.config_setting_group(
        name = "1_or_default",
        match_any = [":condition1", "//conditions:default"],
    )

Compare each element instead. An ORing group containing
`"//conditions:default"` now reduces to an always-true condition, and an
ANDing group ignores the entry, as originally intended.

### DIFF
--- a/lib/selects.bzl
+++ b/lib/selects.bzl
@@ -138,7 +138,7 @@ def _remove_default_condition(settings):
     """Returns settings with "//conditions:default" entries filtered out."""
     new_settings = []
     for setting in settings:
-        if settings != "//conditions:default":
+        if setting != "//conditions:default":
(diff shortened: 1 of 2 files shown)


fmeum (approved):
@gregestren

gregestren (approved):
Wow, what a bug!

eric-wieser (approved):

tetromino (approved):

PR #638, opened 2026-09-01
Rename .bzl files to .scl where possible

eric-wieser:
This makes it possible to use skylib in other starlark environments like copybara.

This leaves behind a `.bzl` wrapper for every `scl` file so as not to break any downstream imports.

`types.scl` has special handling here; it is split in two, since all the types except `depset` are available in scl files.

This diff is rather hard to review due to the compatibility redirects; the SCL files can be understood with:
```
$ for f in lib/*.scl; do   b="lib/$(basename "$f" .scl).bzl";   echo "==================== $b -> $f
====================";   git diff HEAD~1:"$b" "$f"; done
```
```diff
==================== lib/collections.bzl -> lib/collections.scl ====================
==================== lib/dicts.bzl -> lib/dicts.scl ====================
==================== lib/new_sets.bzl -> lib/new_sets.scl ====================
diff --git a/lib/new_sets.bzl b/lib/new_sets.scl
index cd90a30..4af3bbe 100644
--- a/lib/new_sets.bzl
+++ b/lib/new_sets.scl
@@ -23,7 +23,7 @@
   `types.is_set()` method in types.bzl.
 """
 
-load(":dicts.bzl", "dicts")
+load("//lib:dicts.scl", "dicts")
 
 def _make(elements = None):
     """Creates a new set.
==================== lib/partial.bzl -> lib/partial.scl ====================
==================== lib/paths.bzl -> lib/paths.scl ====================
==================== lib/sets.bzl -> lib/sets.scl ====================
diff --git a/lib/sets.bzl b/lib/sets.scl
index 448c498..0aefc14 100644
--- a/lib/sets.bzl
+++ b/lib/sets.scl
@@ -14,6 +14,6 @@
 
 """Skylib module reexporting deprecated set algorithms."""
 
-load(":new_sets.bzl", _sets = "sets")
+load("//lib:new_sets.scl", _sets = "sets")
 
 sets = _sets
==================== lib/shell.bzl -> lib/shell.scl ====================
==================== lib/structs.bzl -> lib/structs.scl ====================
==================== lib/types.bzl -> lib/types.scl ====================
```


tetromino:
This will break Bazel 7 compatibility. @brandjon @susinmotion - should we do it?

susinmotion:
> This will break Bazel 7 compatibility. @brandjon @susinmotion - should we do it?

This sounds fine to me. @meteorcloudy do you have an opinion?

eric-wieser:
> This makes it possible to use skylib in other starlark environments like copybara.

Note that as of writing this is not strictly true, there is a change I have to copybara that will need to land too in order to support `load("//lib:new_sets.scl", ...)` syntax (with a `:` instead of `/`).

tetromino:
Out of curiosity, why would you want skylib's `new_sets` in Copybara? IIRC Copybara's interpreter is new enough to have the native Starlark set data type

eric-wieser:
@tetromino, It's not so much that I want any of these files directly, but that I want to include another `.bzl` file that uses them, and this file can't be converted to `.scl` until its dependencies in skylib are. I gave `new_sets` as an example only because it appears in this diff.

eric-wieser:
> there is a change I have to copybara that will need to land too in order to support `load("//lib:new_sets.scl", ...)` syntax (with a `:` instead of `/`).

This landed in https://github.com/google/copybara/commit/6be2996fd2c94aefdbdf0a4036ec07be1db61ac0

